PR #11, opened 2017-04-15
sample fixes and one more sample

JuaneloJuanelo:


samples/word/01-basics/basic-api-call.yaml
```diff
@@ -13,13 +12,10 @@ script:
         async function run() {
             try {
                 await Word.run(async (context) => {
-                    const range = context.document.getSelection();
+                    let range = context.document.getSelection();
```
Zlatkovsky: `const` is better practice in TS.  It's a stronger version of `let`, and has helped me catch real-world bugs before.  https://medium.com/javascript-scene/javascript-es6-var-let-or-const-ba58b8dcde75

samples/word/01-basics/basic-api-call.yaml
```diff
@@ -1,7 +1,6 @@
-order: 1
-id: word-basic-api-call
+id: word-basics-basic-api-call-
 name: Basic API call
-description: Executes a basic Word API call
+description: Colors the selected text in red.
```
Zlatkovsky: The reason for keeping it generic is so that when folks click on this basic sample, they don't get a bunch of "colors the selected text in red" descriptions in their own snippets (we copy over the descriptions from imported snippets / samples)

samples/word/01-basics/basic-api-call.yaml
```diff
@@ -1,7 +1,6 @@
-order: 1
```
Zlatkovsky: You probably want the order (unless you have some other means by which you want to see these sorted)

samples/word/01-basics/basic-doc-assembly.yaml
```diff
@@ -72,19 +68,17 @@ script:
                 var results = context.document.body.search("Contractor");
                 context.load(results);
                 return context.sync().then(function () {
-                // Once we have the results, we iterate  
-                for (var i = 0; i < results.items.length; i++) {
+                    // Once we have the results, we iterate
```
Zlatkovsky: You can keep this comment if you wish, but it does seem pretty self-explanatory

samples/word/01-basics/basic-doc-assembly.yaml
```diff
@@ -100,15 +94,13 @@ script:
...
-                    })
-                });
+                    .catch(OfficeHelpers.Utilities.log);
+            });
         }
 
         function addFooter() {
             Word.run(function (context) {
-            // Create a proxy collection for the sections collection.
+                // Create a proxy collection for the sections collection.
```
Zlatkovsky: At least on the Excel side, we've removed most of the "creates a proxy object..." comments

samples/word/01-basics/insert-and-get-pictures.yaml
```diff
@@ -1,4 +1,4 @@
-id: word-basics-insert-and-get-pictures
+id: word-basics-insert-and-get-pictures-
```
Zlatkovsky: is it actually my tool that adds all these dashes?  if so, don't worry about fixing them, I'll fix it via the tool itself.

samples/word/01-basics/search.yaml
```diff
@@ -25,39 +25,36 @@ script:
...
+                .catch(OfficeHelpers.Utilities.log);
         }
 
-        function setup() { 
+        function setup() {
             Word.run(function (context) {
                 // lets insert a couple of paragraphs to illustrate the point..
                 context.document.body.clear();
-                context.document.body.insertParagraph("Video provides a powerful way to help you prove  your point. When you click Online Video, you can paste in the embed code for the video you want to add. You can also type a keyword to search online for the video that best fits your document.", "start"); 
+                context.document.body.insertParagraph("Video provides a powerful way to help you prove your point. When you click Online Video ($10,000.00), you can paste in the embed code for the video you want to add. You can also type a keyword to search online for the video that best fits your document.", "start");
```
Zlatkovsky: Where did this come from?  Accidental paste? :-)

samples/word/04-range/split-words-of-first-paragraph.yaml
```diff
@@ -0,0 +1,161 @@
...
+author: Microsoft
+host: WORD
+api_set:
+    WordApi: 1.1
+script:
+    content: |-
+        $("#setup").click(setup);
+        $("#highlight").click(highlightWords);
+
+        function highlightWords() {
```
Zlatkovsky: This sample would seriously benefit from TypeScript-ification.  And you'd actually be able to get rid of the entirety of the nesting

samples/word/04-range/split-words-of-first-paragraph.yaml
```diff
@@ -0,0 +1,161 @@
...
+            })
+                .catch(OfficeHelpers.Utilities.log);
+
+            function createTimerPromise(ms) {
+                return new OfficeExtension.Promise(function (resolve) {
+                    setTimeout(resolve, ms);
+                })
+            }
+
+            function forEach(collection, handler) {
```
Zlatkovsky: Wouldn't need this if you were using TS... Just sayin' :-)

samples/word/04-range/split-words-of-first-paragraph.yaml
```diff
@@ -0,0 +1,161 @@
...
+                var promise = new OfficeExtension.Promise(function (resolve) { resolve(); });
```
Zlatkovsky: BTW, this code will fail on RTM, where we don't (yet) expose OfficeExtension.Promise.  you can use a regular "Promise" instead, though.  Or use async/await, which uses "Promise" under the covers.

samples/word/04-range/split-words-of-first-paragraph.yaml
```diff
@@ -0,0 +1,161 @@
...
+    language: html
+style:
+    content: |4
+            body {
+                margin: 0;
+                padding: 10px;
+            }
+
+
+            /* Button customization, including overwriting some Fabric defaults */
```
Zlatkovsky: Here and elsewhere:  we've essentially cut out all of the CSS, as it's not really needed except for the coloring, and adds a bunch of bloat that Jakob felt was unnecessary

samples/word/01-basics/basic-api-call.yaml
Zlatkovsky: FWIW, esp. on typescriptified (async/await) samples, I would definitely use "const" everywhere I can, "let" everywhere else, and "var" never.  So when you have the time, it'd be great if you can make those changes in the other samples as well.

samples/word/07-custom-properties/custom-properties-access.yaml
```diff
@@ -1,24 +1,26 @@
...
-                         })
-                 })
-            }
+        function getProperties() {
+            Word.run(function (context) {
+                var customDocumentProperties = context.document.properties.customProperties;
+                context.load(customDocumentProperties);
+                return context.sync()
+                    .then(function () {
+                        console.log(customDocumentProperties.items.length);
```
Zlatkovsky: Do you want to list them out (console.log of the items) instead of just the length?

samples/word/07-custom-properties/custom-properties-access.yaml
```diff
@@ -1,24 +1,26 @@
...
-                    return context.sync()
-                        .then(function () {
-                            console.log(customDocumentProperties.items.length);
-                         })
-                 })
-            }
+        function getProperties() {
+            Word.run(function (context) {
+                var customDocumentProperties = context.document.properties.customProperties;
+                context.load(customDocumentProperties);
```
Zlatkovsky: Here an elsewhere, not specifying the property names you want is BAD PRACTICE.  More about it in Chapter 5 of my book, which you should have access to (if not, email me, and I'll send you a link for a free copy for our team).

samples/word/04-range/split-words-of-first-paragraph.yaml
```diff
@@ -0,0 +1,161 @@
...
+script:
+    content: |-
+        $("#setup").click(setup);
+        $("#highlight").click(highlightWords);
+
+        function highlightWords() {
+            Word.run(function (context) {
+                // var myWords = context.document.body.paragraphs.first.split([" "], true, true);
+                var myPars = context.document.body.paragraphs;
+                context.load(myPars, { expand: 'font' });
```
Zlatkovsky: Why are you doing expand?  I have yet to see a place where it was actually needed, and a good idea.  Doing a regular load and only of the properties you need would be far better.

samples/word/04-range/split-words-of-first-paragraph.yaml
```diff
@@ -0,0 +1,161 @@
...
+                return context.sync()
+                    .then(function () {
+                        var myWords = myPars.items[0].split([" "], true, true);
```
Zlatkovsky: Nit: would be nice to specify what those "true" arguments are for.

Eg.,:

    items[0].split([" "], true /*matchCase*/, true /*soundsLike*/)

or whatever the parameters are (made them up above)

samples/word/07-custom-properties/get-built-in-properties.yaml
```diff
@@ -0,0 +1,96 @@
...
+script:
+    content: |+
+        $("#run").click(getProperties);
+
+        function getProperties() {
+            Word.run(function (context) {
+                var builtInProperties = context.document.properties;
+                context.load(builtInProperties);
+                return context.sync()
+                    .then(function () {
```
Zlatkovsky: You could just do JSON.stringify(builtInProperties, null, 4) and get them all at once...

samples/word/07-custom-properties/read-write-custom-document-properties.yaml
```diff
@@ -11,45 +11,39 @@ script:
...
                 var properties = context.document.properties.customProperties;
                 context.load(properties);
                 return context.sync()
                     .then(function () {
                         for (var i = 0; i < properties.items.length; i++)
-                            console.log("Property Name:" + properties.items[i].key + ";Type=" + properties.items[i].type +"; Property Value=" + properties.items[i].value);
-                    })
-                    .catch(function (e) { 
-                        console.log(e.message);
+                            console.log("Property Name:" + properties.items[i].key + ";Type=" + properties.items[i].type + "; Property Value=" + properties.items[i].value);
```
Zlatkovsky: Nit, add spaces.
You would also be far better off with a TypeScript template string https://basarat.gitbooks.io/typescript/content/docs/template-strings.html

samples/word/01-basics/basic-api-call.yaml
```diff
@@ -13,13 +12,10 @@ script:
         async function run() {
             try {
                 await Word.run(async (context) => {
-                    const range = context.document.getSelection();
+                    let range = context.document.getSelection();
                     range.font.color = "red";
-                    range.load("text");
```
Zlatkovsky: I think it's useful showing the load pattern as well.  Esp. to contrast with the Promise-based JS-only version

samples/word/01-basics/basic-api-call.yaml
jakobpn: It would be great if we could get the TS basic api call at the top :) Also, I wonder if we should make the Basics folder smaller and move some of the snippets to other folders?

samples/word/01-basics/basic-api-call.yaml
jakobpn: We made this change throughout the Excel snippets.

samples/word/04-range/split-words-of-first-paragraph.yaml
jakobpn: Juan, I will send you a template that you can use. Or you can check for the HTML and CSS in the Excel snippets.

samples/word/01-basics/basic-api-call.yaml
JuaneloJuanelo: i already moved a bunch of the samples to other folders. The samples that are on basics, are really the basics. I will adjust the order of some of them.

samples/word/01-basics/basic-api-call.yaml
JuaneloJuanelo: not sure whats the feedback here. i not even created this sample! :)

samples/word/01-basics/basic-api-call.yaml
JuaneloJuanelo: i did not wrote this sample but will fix this...

samples/word/01-basics/basic-doc-assembly.yaml
JuaneloJuanelo: ok. ill keep it. :)

samples/word/01-basics/basic-api-call.yaml
JuaneloJuanelo: fixed

samples/word/01-basics/basic-api-call.yaml
JuaneloJuanelo: reverted the description. local fixed

samples/word/01-basics/basic-doc-assembly.yaml
JuaneloJuanelo: local fixed

samples/word/01-basics/insert-and-get-pictures.yaml
JuaneloJuanelo: yes ... the tool added the ids.

samples/word/01-basics/search.yaml
JuaneloJuanelo: no i needed to add a currency formatted text on the sample text, so the search by wildcard example can find it,.

samples/word/04-range/split-words-of-first-paragraph.yaml
JuaneloJuanelo: ok, for now i will keep it like this. i will typescript it later if time permits.

samples/word/04-range/split-words-of-first-paragraph.yaml
JuaneloJuanelo: yeap, good catch this is not neeed. in a previous example i was getting some info on the font, but forgot to remove the load.
local fixed.

samples/word/04-range/split-words-of-first-paragraph.yaml
JuaneloJuanelo: local fixed

samples/word/04-range/split-words-of-first-paragraph.yaml
JuaneloJuanelo: :) yeap can wait to learn it, in the meantime i will keep it like this.

samples/word/04-range/split-words-of-first-paragraph.yaml
JuaneloJuanelo: ok

samples/word/04-range/split-words-of-first-paragraph.yaml
JuaneloJuanelo: i see this as a p3 change. i will prioritize adding more samples and will fix it later (it requires changing all the samples) also it should be ok to use CSS.

samples/word/07-custom-properties/custom-properties-access.yaml
JuaneloJuanelo: in this example i want to retrieve all.

samples/word/07-custom-properties/custom-properties-access.yaml
JuaneloJuanelo: yeap i think i will just leave the other example, who actually shows how to create and retrieve custom doc props. this is basically a dupe...

samples/word/07-custom-properties/get-built-in-properties.yaml
JuaneloJuanelo: great... local fixed....

samples/word/07-custom-properties/read-write-custom-document-properties.yaml
JuaneloJuanelo: for now I will leave it as is.

PR #21, opened 2017-04-28
Create selection-changed.xaml

jakobpn:


samples/excel/09-events/selection-changed.yaml
```diff
@@ -0,0 +1,116 @@
...
+        $("#add-event-handler").click(addEventHandler);
+        $("#remove-last-event-handler").click(removeLastEventHandler);
+        $("#remove-all-event-handlers").click(removeAllEventHandlers);
+
+        let lastHandler: OfficeExtension.EventHandlerResult<Excel.Workbook>;
+        let workbook: Excel.Workbook;
+
+        async function addEventHandler() {
+            try {
+                if (workbook !== undefined) {
```
Zlatkovsky: FYI:  There is a pending bug fix on my plate that will make this workbook !== undefined not be needed.  I.e., it will make Excel.run be OK with taking in a null/undefined object.  So in a few weeks, we should be able to remove this.

samples/excel/09-events/selection-changed.yaml
```diff
@@ -0,0 +1,116 @@
...
+                    await Excel.run(workbook, async (context) => {
+                        workbook = context.workbook;
+                        lastHandler = context.workbook.onSelectionChanged.add(onSelectionChanged)
```
Zlatkovsky: Missing semicolon.  Ditto below.

samples/excel/09-events/selection-changed.yaml
```diff
@@ -0,0 +1,116 @@
...
+
+                        await context.sync();
+                    })
+                } else {
```
Zlatkovsky: An else to an "!==" if statement implies that the blocks of code should be reversed.

samples/excel/09-events/selection-changed.yaml
Zlatkovsky: Just because we have this bug, though, doesn't mean you need to duplicate the code entirely.  Why don't you make the body (the batch) into its own inlined function below, and that wait it's only the Excel.run call that is different?

	async function addEventHandler() {
		try {
			if (workbook === undefined) {
				await Excel.run(addHandlerHelper);
			} else {
				await Excel.run(workbook, addHandlerHelper);
			}
			console.log("Event handler added. Try changing the selection.");
		}
		catch (error) {
			OfficeHelpers.Utilities.log(error);
		}
		

		async function addHandlerHelper(context: Excel.RequestContext) {
			workbook = context.workbook;
			lastHandler = context.workbook.onSelectionChanged.add(onSelectionChanged)

			await context.sync();
		});
	}

samples/excel/09-events/selection-changed.yaml
```diff
@@ -0,0 +1,116 @@
...
+                    console.log("All event handlers removed");
+                })
+            }
+            catch (error) {
+                OfficeHelpers.Utilities.log(error);
+            }
+        }
+
+        async function onSelectionChanged() {
+            console.log("Selection changed, let's look up the address");
```
Zlatkovsky: I didn't literally mean you to have this console.log line :-).  It could probably be merged in to where you actually fetch the address, rather than having two separate lines.  Unless you intentionally wanted this?...

samples/excel/09-events/selection-changed.yaml
```diff
@@ -0,0 +1,116 @@
+name: Selection Changed
+description: Add and remove an event handler on the selection changed event
+host: EXCEL
+api_set: {}
```
Zlatkovsky: ApiSet: ExcelApi 1.2

samples/excel/09-events/selection-changed.yaml
jakobpn: Good change.

jakobpn: Great suggestion. Fixed in the next commit.

samples/excel/09-events/selection-changed.yaml
jakobpn: Fixed.

samples/excel/09-events/selection-changed.yaml
jakobpn: Fixed in next commit.

samples/excel/09-events/selection-changed.yaml
jakobpn: Fixed in the next commit.

samples/excel/09-events/selection-changed.yaml
jakobpn: Fixed

Zlatkovsky:
Looks great!  Let me go ahead and merge it in.

PR #23, opened 2017-05-04
Create conditional-formatting.yaml

jakobpn:


Zlatkovsky:
@jakobpn , I don't want to merge this in quite yet, until we can hide snippets that are inapplicable for a particular host.  See https://github.com/OfficeDev/script-lab/issues/305.  We're likely ~2 months away before the general public gets to use conditional formatting.

Once we've done that work to hide inapplicable snippets, I can merge this one in.

Zlatkovsky:
@jakobpn , I wasn't able to see it before, since the code didn't paste in a parseable way in the pull request -- but there is something definitely wrong in this section:

            range.conditionalFormats.add(Excel.ConditionalFormatType.iconSet);

            const conditionalFormatCount = range.conditionalFormats.getCount();

            await context.sync();

            const conditionalFormat = range.conditionalFormats.getItemAt(conditionalFormatCount.value - 1);

            context.load(conditionalFormat);

            conditionalFormat.priority = 0;

            await context.sync();


Namely:

- Why are you re-fetching the item (and using "count"... both overly complex and unsafe) rather than storing it in a variable and using it immediately?
- Why are you doing a `context.load(conditionalFormat)`?  What is it that you're trying to get out of it, and why are you not specify the properties you want (if you actually do want them)?
- Why the two extra `context.sync`-s? -- I don't see an immediate need for either of them.

Zlatkovsky:
For the click handler, a couple issues with that as well that are worth fixing:

                       .click({ worksheet: sheet.name, index: i }, async (e: any) => {
                            e.preventDefault();
                            const sheet = context.workbook.worksheets.getItem(e.data.worksheet);
                            const conditionalFormat = sheet.getRange()
                                .conditionalFormats.getItemAt(e.data.index);
                            const range = conditionalFormat.getRange();
                            range.select();
                            await context.sync();

                        });

1. You should [probably] use the `Excel.run(previousObject, batch)` syntax -- just for the consistency of always using `Excel.run` when creating a new batch of work.
2. You need error handling

Zlatkovsky:
Nit: I would also create a reference to the CF object directly rather than always directing via the worksheetRange.  Just for less typing:

            const conditionalFormats = sheet.getRange().conditionalFormats.load("type")

            await context.sync();

            $("#conditional-formats li").remove();

            if (conditionalFormats.items.length > 0) {
                for (let i = 0; i < conditionalFormats.items.length; i++) {
                    let $a = $(`<a href=''>${conditionalFormats.items[i].type}</a>`)

Zlatkovsky:
Why `sheet.load("name")`?

Zlatkovsky:
For IntelliSense purposes, doing `const criteria = { ... }` does not give you IntelliSense, whereas setting it inline on the conditional formatting object would.  Please consider changing to something like:

            conditionalFormat.colorScale.criteria = {
				minimum: { formula: null, type: Excel.ConditionalFormatColorCriterionType.lowestValue, color: "blue" },
				midpoint: { formula: "50", type: Excel.ConditionalFormatColorCriterionType.percent, color: "yellow" },
				maximum: { formula: null, type: Excel.ConditionalFormatColorCriterionType.highestValue, color: "red" }
            };

jakobpn:
Thx for the review comments Michael. I committed a new version.

Rick-Kirkham:
This snippet was superseded by an expanded version: https://github.com/OfficeDev/office-js-snippets/pull/66

PR #30, opened 2017-05-25
Added ColorWheel, Gradient, and Patterns

AlexanderZlatkovski:


samples/excel/90-just-for-fun/color-wheel.yaml
```diff
@@ -0,0 +1,144 @@
...
+                            matrix[r][c] = 1;
+                        }
+                    }
+                    dataRange.values = matrix;
+
+                    // Create a sheet for the color wheel and format it
+                    const sheet = await OfficeHelpers.ExcelUtilities.forceCreateSheet(context.workbook, "Color Wheel", true /*clearOnly*/);
+                    
+                    sheet.charts.load("id");
+                    sheet.activate();
```
jakobpn: sheet.activate() need to be done in with a new context.sync() after the batch that creates the worksheet. This is a workaround to an issue that we discovered with Excel Online.

jakobpn: Creating a new sheet and activating it with the same context.sync() will cause an issue in Excel Online. Insert a context.sync() right before sheet.activate().

samples/excel/90-just-for-fun/gradient.yaml
```diff
@@ -0,0 +1,181 @@
...
+            let range = sheet.getCell(1, 1).getResizedRange(
+                originalSize - 1, originalSize - 1);
+            for (let row = 0; row < originalSize; row++) {
+                for (let col = 0; col < originalSize; col++) {
+                    range.getCell(row, col).format.fill.color =
+                        colors2D[row][col];
+                }
+            }
+
+            sheet.activate();
```
jakobpn: Creating a new sheet and activating it with the same context.sync() will cause an issue in Excel Online. Insert a context.sync() right before sheet.activate().

samples/excel/90-just-for-fun/patterns.yaml
```diff
@@ -0,0 +1,210 @@
...
+    content: |-
+        $("#spiral").click(drawSpiral);
+        $("#squares").click(drawSquares);
+        $("#cross").click(drawCross);
+
+        async function drawSquares() {
+            try {
+                await Excel.run(async (context) => {
+                    const sheet = await OfficeHelpers.ExcelUtilities.forceCreateSheet(context.workbook, "Patterns", true /*clearOnly*/);
+                    sheet.activate();
```
jakobpn: Creating a new sheet and activating it with the same context.sync() will cause an issue in Excel Online. Insert a context.sync() right before sheet.activate().

Zlatkovsky: @jakobpn , I think we can do a fix in the `OfficeHelpers` instead.  Let me talk with you offline.

Zlatkovsky: @jakobpn , created pull request https://github.com/OfficeDev/office-js-helpers/pull/30.
Once approve and checked in, the samples can go in as is -- except we'll need to bump the version number of OfficeJsHelpers from 0.7.1 to 0.7.2.  However, I think we want to do it all-up for all samples.  So my suggestion is that once OfficeJsHelpers is patched, you approve and merge in this pull request, and then I can follow up with a different pull request that updates all the samples to 0.7.2 in bulk.

Zlatkovsky: In fact, at that point, I think we can just say that the fix to Office Helpers is orthogonal to these samples (and we already have the exact same problem for all of our other samples anyway).

@jakobpn , could you go ahead and merge this in?  And I'll do the separate PR to update all samples to 0.7.2.  Filing bug https://github.com/OfficeDev/office-js-snippets/issues/31 to track the latter.

PR #38, opened 2017-07-10
adding snippet for for Rich API object.set method

Rick-Kirkham:
In addition to adding a reference to my new snippet in excel.yaml, it looks like **npm start** deleted a snippet reference from the same file. I don't know why, but the instructions say run **npm start** and then submit, so that's what I'm doing.


playlists/excel.yaml
```diff
@@ -20,6 +20,17 @@
   group: Basics
```
jakobpn: I don't believe this file should be part of the commit.

samples/excel/01-basics/multiple-property-set.yaml
```diff
@@ -0,0 +1,156 @@
...
+description: Setting multiple properties at once with the rich API object set() method.
+author: OfficeDev
+host: EXCEL
+api_set:
+    ExcelApi: 1.4
+script:
+    content: |
+        $("#setup").click(setup);
+
+                $("#set-multiple-properties-with-object").click(setMultiplePropertiesWithObject);
```
jakobpn: Please remove the indentation starting from this line. Will be easier to read the code in the narrow task pane.

samples/excel/01-basics/multiple-property-set.yaml
```diff
@@ -0,0 +1,156 @@
...
+                    try {
+                        await Excel.run(async (context) => {
+                            const sheet = context.workbook.worksheets.getItem("Sample");
+
+                            const sourceRange = sheet.getRange("B2:E2");
+                            sourceRange.load("format/fill/color, format/font/name, format/font/color");
+                            await context.sync();
+
+                            const targetRange = sheet.getRange("B7:E7");
+                            console.log(sourceRange.format.font.color);
```
jakobpn: Not sure what the purpose of this console.log is. I would recommend removing it.

playlists/excel.yaml
Zlatkovsky: No, go ahead and commit it.  I think it's useful to see for diffing's sake, and there is no harm (Travis will just override it anyway)

samples/excel/01-basics/multiple-property-set.yaml
```diff
@@ -0,0 +1,155 @@
...
+			}
+		}
+
+		async function copyPropertiesFromRange() {
+			try {
+				await Excel.run(async (context) => {
+					const sheet = context.workbook.worksheets.getItem("Sample");
+
+					const sourceRange = sheet.getRange("B2:E2");
+					sourceRange.load("format/fill/color, format/font/name, format/font/color");
```
Zlatkovsky: Why are you loading these properties, if you're not reading them back?

samples/excel/01-basics/multiple-property-set.yaml
```diff
@@ -0,0 +1,155 @@
...
+		$("#copy-properties-from-range").click(copyPropertiesFromRange);
+
+		async function setMultiplePropertiesWithObject() {
+			try {
+				await Excel.run(async (context) => {
+					const sheet = context.workbook.worksheets.getItem("Sample");
+
+					const range = sheet.getRange("B2:E2");
+					range.set({
+						format: {
```
Zlatkovsky: I feel like this sample would be more interesting if you also set a few other properties.  Like values and number formats.

samples/excel/01-basics/multiple-property-set.yaml
```diff
@@ -0,0 +1,155 @@
...
+			try {
+				await Excel.run(async (context) => {
+					const sheet = context.workbook.worksheets.getItem("Sample");
+
+					const sourceRange = sheet.getRange("B2:E2");
+					sourceRange.load("format/fill/color, format/font/name, format/font/color");
+					await context.sync();
+
+					const targetRange = sheet.getRange("B7:E7");
+					targetRange.set(sourceRange);
```
Zlatkovsky: It may be worth, via a comment, to highlight how you are now setting object properties based on the loaded-and-synced source range.

samples/excel/01-basics/multiple-property-set.yaml
Zlatkovsky: ... Well, ok, maybe not values (I looked at the setup code now, and understand the scenario better).  But maybe number formats?

Rick-Kirkham: Well, it currently is designed to change only the header range. Do you want an example that changes the styling and number format of the data range?

Zlatkovsky: I just think it would be good to show that it's not just nested properties (format/font and format/fill) that you can set, but also a root-level property (like values)

Charts is another good object to do a "set" on, it has a bunch of properties at both the root level and deeper-nested (and "set" makes such API calls look much cleaner than having a wall of repeating text)

Zlatkovsky:
Looks good to me!  Will merge as soon as Travis finishes building

Issue #29, opened 2017-05-22
Is "const" in snippets causing confusion?  Should we have done only "let"-s?


Zlatkovsky:
@jakobpn , would be curious to hear your opinion on this.  Internally and externally, I feel like folks have had a hard time reasoning about what a `const sheet =` is (e.g., thinking that it's a protected sheet that you can't add to, etc.).

Should we just bulk-change all `const`-s to `let`-s?

jakobpn:
I think we should take advantage of "const" providing a safety net that help ensure that if you don't change a reference to an object by mistake. My vote is to keep the snippets as is.

Zlatkovsky:
Agreed in principle.  In practice, though, I have yet to see it catch an error for me.  Whereas I have seen it cause confusion...

I'm split on this (for my own code, I would use `const` -- but I'm actually not sure that I would when explaining in, say, my book).  So as you interact with customers, would be curious to hear what their impressions are.

jakobpn:
So far it I don't think I have observed confusion, but it is something that I will look after.

Rick-Kirkham:
FWIW, I find it puzzling. In every other language I know, `const` is used for something whose value is known and assigned by the developer as a literal at coding time.  It is weird to see a `const `initialized at runtime with the return value of a method.

AlexJerabek:
@Zlatkovsky and @Rick-Kirkham , I'm curious if you've had further thoughts on this matter a year and a half later. My thoughts are that tslint enforces const where possible, so we should try to conform to what's ostensibly our own (MSFT) guidance.

Rick-Kirkham:
@AlexJerabek @Zlatkovsky I've reversed myself. I now think we should use const as it is intended in TypeScript. It is initially confusing to people who know other languages with "const" because it has a different meaning in TS, but I think now that that's just part of the process of learning a new language.

Issue #28, opened 2017-05-17
Remove many of the `console.log`-s and replace with OfficeHelpers.UI.notify


Zlatkovsky:
Example: dates snippet, when doing a "get"

Zlatkovsky:
I only did for the handful of snippets I touched, but not for the vast majority of them (including the other hosts)

kbrandl:
@AlexJerabek "OfficeHelpers" is publicly available -- any user can get use it by including the [@microsoft/office-js-helpers](https://www.npmjs.com/package/@microsoft/office-js-helpers) package in their project (source repo in GitHub is [OfficeDev/office-js-helpers](https://github.com/OfficeDev/office-js-helpers)).  In Script Lab, a snippet gains access to it by virtue of including it within the list of **Libraries** that the snippet specifies.

That said though, I tend to agree with you -- i.e., I prefer using `console.log()` over using `OfficeHelpers.UI.notify` for a few reasons:
- Like you, I have trouble noticing messages that `OfficeHelpers.UI.notify` displays, since it overlays directly over top of task pane and doesn't have any visual features that make it stand out.
- You can only display a single `OfficeHelpers.UI.notify` message at a time; each message overwrites the last (while `console.log` allows you to see the full list of messages that've been written until if/when you clear the console -- more effective for seeing the flow of what's occurred, which is what developers are usually most interested in).
- If a developer copies a function out of Script Lab into their add-in project and does not also include the reference to the **office-js-helpers** package, they'll get an error (if the function uses `OfficeHelpers`).

There are arguments to be made *for* using `OfficeHelpers` as well, but personally I prefer `console.log`...just my 2 cents.

Zlatkovsky:
That's fine, I think you can close the issue and leave things be.
The original concern was that for a lot of people, I saw that they minimize Firebug once, and thus don't see any console output.  However, in the reworked version of Script Lab, we will have our own console component (for accessibility reasons, among others), and so there we can just always make sure to pop it open if output comes out.   So console.log is fine.
Feel free to close as "won't fix"

PR #42, opened 2017-07-12
Snippet for making a named item from a range

Rick-Kirkham:
As expected, `npm start` added a reference to the new snippet to excel.yaml. It also deleted blank lines from some other files. I don't know how those blank lines got there. I didn't ever have those files open.

### DIFF
--- a/samples/excel/01-basics/basic-api-call.yaml
+++ b/samples/excel/01-basics/basic-api-call.yaml
@@ -32,7 +32,6 @@ script:
                 OfficeHelpers.Utilities.log(error);
             }
         }
-
     language: typescript
 template:
     content: |
--- a/samples/excel/30-range/test-for-used-range.yaml
+++ b/samples/excel/30-range/test-for-used-range.yaml
@@ -1,0 +1,125 @@
+id: excel-test-for-used-range
+name: Test for used range
+description: Create a chart from a table only if there's data in the table.
+author: OfficeDev
+host: EXCEL
+api_set:
+    ExcelApi: 1.4
+script:
+    content: |
+        $("#setup").click(() => tryCatch(setup));
+        $("#try-create-chart-from-empty-table").click(() => tryCatch(tryCreateChartFromEmptyTable));
+
+        async function tryCreateChartFromEmptyTable() {
+            await Excel.run(async (context) => {
+
+                const sheet = context.workbook.worksheets.getItem("Sample");    
+                const salesTable = sheet.tables.getItem("SalesTable");
+                const dataRange = salesTable.getDataBodyRange();
+                const usedDataRange = dataRange.getUsedRangeOrNullObject(true /* only cells with values count as used */);
+
+                //Must sync before reading value returned from *OrNullObject method/property.
+                await context.sync();
+
+                if (!usedDataRange.isNullObject) {
+                    const chart = sheet.charts.add("ColumnClustered", dataRange, "auto");
+                    chart.setPosition("A15", "F30");
+                    chart.title.text = "Quarterly sales chart";
+                    chart.legend.position = "right"
+                    chart.legend.format.fill.setSolidColor("white");
+                    chart.dataLabels.format.font.size = 15;
+                    chart.dataLabels.format.font.color = "black";
+                } else {            
+                    OfficeHelpers.UI.notify("To create a meaningful chart, add names to the Product column and numbers to some of the other cells. Then press 'Try to create chart' again.");
+                }
+
+                await context.sync();
+            });
+        }
+
+        async function setup() {
+            await Excel.run(async (context) => {
+                await OfficeHelpers.ExcelUtilities.forceCreateSheet(context.workbook, "Sample");
+                const sheet = context.workbook.worksheets.getItem("Sample");
+                let salesTable = sheet.tables.add('B2:F2', true);         
+                salesTable.name = "SalesTable";
+                salesTable.showTotals = true;
+                salesTable.getHeaderRowRange().values = [["Product", "Qtr1", "Qtr2", "Qtr3", "Qtr4"]];
+
+                // The table should be created wtih no data.
+                salesTable.rows.add(null, [
+                    [null, null, null, null, null],
+                    [null, null, null, null, null],
+                    [null, null, null, null, null],
+                    [null, null, null, null, null],
+                    [null, null, null, null, null],
+                    [null, null, null, null, null]
+                ]);  
+                sheet.activate();
+
+                await context.sync();
+            });
+        }
+
+        /** Default helper for invoking an action and handling errors. */
+        async function tryCatch(callback) {
+            try {
+                await callback();
+            }
+            catch (error) {
+                OfficeHelpers.UI.notify(error);
+                OfficeHelpers.Utilities.log(error);
+            }
+        }
+    language: typescript
+template:
+    content: |
+        <section class="ms-font-m">
+            <p>This sample shows how to use the getUsedRangeOrNullObject() function.</p>
+        </section>
+
+        <section class="setup ms-font-m">
+            <h3>Set up</h3>
+            <button id="setup" class="ms-Button">
+                <span class="ms-Button-label">Add sample data</span>
+            </button>
+        </section>
+
+        <section class="samples ms-font-m">
+            <h3>Try it out</h3>
+            <button id="try-create-chart-from-empty-table" class="ms-Button">
+                <span class="ms-Button-label">Try to create chart</span>
+            </button>   
+        </section>
+    language: html
+style:
+    content: |
+        section.samples {
+            margin-top: 20px;
+        }
+
+        section.samples .ms-Button, section.setup .ms-Button {
+            display: block;
+            margin-bottom: 5px;
+            margin-left: 20px;
+            min-width: 80px;
+        }
+    language: css
+libraries: |
+    # Office.js
+    https://appsforoffice.microsoft.com/lib/1/hosted/office.js
+
+    # CSS Libraries
+    office-ui-fabric-js@1.4.0/dist/css/fabric.min.css
+    office-ui-fabric-js@1.4.0/dist/css/fabric.components.min.css
+
+    # NPM libraries
+    core-js@2.4.1/client/core.min.js
+    @microsoft/office-js-helpers@0.7.4/dist/office.helpers.min.js
+    jquery@3.1.1
+
+    # IntelliSense: @types/library or node_modules paths or URL to d.ts files
+    @types/office-js
+    @types/core-js
+    @microsoft/office-js-helpers@0.7.4/dist/office.helpers.d.ts
+    @types/jquery
--- a/samples/excel/45-named-item/create-and-use-named-item-for-range.yaml
+++ b/samples/excel/45-named-item/create-and-use-named-item-for-range.yaml
@@ -1,0 +1,180 @@
+id: excel-create-and-use-named-range-item
+name: Create and use named range item
+description: Create and use a named range item
+author: OfficeDev
+host: EXCEL
+api_set:
+    ExcelApi: 1.4
+script:
+    content: |-
+        $("#add-name").click(() => tryCatch(addName));
+        $("#format-named-range").click(() => tryCatch(formatRangeByName));
+        $("#setup").click(() => tryCatch(setup));
+
+        async function addName() {
+            await Excel.run(async (context) => {
+
+                const sheet = context.workbook.worksheets.getItem("Sample");
+                const headerRange = sheet.getRange("A1:E1");
+
+                sheet.names.add("ExpensesHeader", headerRange);
+
+                const namedItems = sheet.names.load();
+
+                await context.sync();
+
+                console.log("This workbook contains " + namedItems.items.length + " named items.");
+
+                for (let i = 0; i < namedItems.items.length; i++) {
+                    console.log(JSON.stringify(namedItems.items[i])) + "\n";
+                }   
+
+                await context.sync();
+            });
+        }
+
+        async function formatRangeByName() {
+            await Excel.run(async (context) => {
+
+                const sheet = context.workbook.worksheets.getItem("Sample");
+
+                const headerRowName = sheet.names.getItem("ExpensesHeader");
+                const headerRange = headerRowName.getRange();
+                headerRange.format.fill.color = "red";
+
+                await context.sync();
+            });
+        }
+
+        async function setup() {
+            await Excel.run(async (context) => {
+
+                let sheet = await OfficeHelpers.ExcelUtilities.forceCreateSheet(context.workbook, "Sample");
+
+                let expensesTable = sheet.tables.add("A1:D1", true);
+                expensesTable.name = "ExpensesTable";
+
+                expensesTable.getHeaderRowRange().values = [["Date", "Merchant", "Category", "Amount"]];
+
+                let newData = transactions.map(item =>
+                    [item.date, item.merchant, item.category, item.amount]);
+
+                expensesTable.rows.add(null, newData);
+                sheet.getUsedRange().format.autofitColumns();
+                sheet.getUsedRange().format.autofitRows();
+                sheet.activate();
+
+                await context.sync();
+            });
+        }
+
+        const transactions = [
+            {
+                "date": "1/1/2017",
+                "merchant": "The Phone Company",
+                "category": "Communications",
+                "amount": "$120"
+            },
+            {
+                "date": "1/1/2017",
+                "merchant": "SouthRidge Video",
+                "category": "Entertainment",
+                "amount": "$40"
+            },
+            {
+                "date": "1/1/2017",
+                "merchant": "Coho Winery",
+                "category": "Restaurant",
+                "amount": "$47"
+            },
+            {
+                "date": "1/2/2017",
+                "merchant": "Contoso, Ltd",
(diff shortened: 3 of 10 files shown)


samples/excel/45-named-item/create-and-use-named-item-for-range.yaml
```diff
@@ -0,0 +1,180 @@
...
+                const sheet = context.workbook.worksheets.getItem("Sample");
+                const headerRange = sheet.getRange("A1:E1");
+
+                sheet.names.add("ExpensesHeader", headerRange);
+
+                const namedItems = sheet.names.load();
+
+                await context.sync();
+
+                console.log("This workbook contains " + namedItems.items.length + " named items.");
```
Zlatkovsky: Please use a template string instead

samples/excel/45-named-item/create-and-use-named-item-for-range.yaml
```diff
@@ -0,0 +1,180 @@
...
+        $("#format-named-range").click(() => tryCatch(formatRangeByName));
+        $("#setup").click(() => tryCatch(setup));
+
+        async function addName() {
+            await Excel.run(async (context) => {
+
+                const sheet = context.workbook.worksheets.getItem("Sample");
+                const headerRange = sheet.getRange("A1:E1");
+
+                sheet.names.add("ExpensesHeader", headerRange);
```
Zlatkovsky: Is there any reason for all the spacing between the lines?  I could probably have no breaks except before and after `context.sync()`

samples/excel/45-named-item/create-and-use-named-item-for-range.yaml
```diff
@@ -0,0 +1,180 @@
...
+
+                const namedItems = sheet.names.load();
```
Zlatkovsky: Ricky, you're the only person who's ready my book cover-to-cover!  Certainly it was only accidental that you omitted the names of the properties that you care about in the `load` statement, right?

image

samples/excel/45-named-item/create-and-use-named-item-for-range.yaml
```diff
@@ -0,0 +1,180 @@
...
+                    console.log(JSON.stringify(namedItems.items[i])) + "\n";
+                }   
+
+                await context.sync();
+            });
+        }
+
+        async function formatRangeByName() {
+            await Excel.run(async (context) => {
+
```
Zlatkovsky: I don't see a need for the extra newline at the beginning of the function (ditto elsewhere)

samples/excel/01-basics/basic-api-call.yaml
```diff
@@ -32,7 +32,6 @@ script:
                 OfficeHelpers.Utilities.log(error);
             }
         }
-
```
Zlatkovsky: Actually, it was me who added these :-)  But that's fine, doesn't really matter whether they're there or not.

samples/excel/45-named-item/create-and-use-named-item-for-range.yaml
```diff
@@ -0,0 +1,180 @@
...
+                sheet.names.add("ExpensesHeader", headerRange);
+
+                const namedItems = sheet.names.load();
+
+                await context.sync();
+
+                console.log("This workbook contains " + namedItems.items.length + " named items.");
+
+                for (let i = 0; i < namedItems.items.length; i++) {
+                    console.log(JSON.stringify(namedItems.items[i])) + "\n";
```
Zlatkovsky: What is the + "\n" doing outside the console log (and you don't need it).
Also, why the JSON.stringify here?

samples/excel/45-named-item/create-and-use-named-item-for-range.yaml
Zlatkovsky: It would be good to start using OfficeHelpers.UI.notify more.  At some point we may try to make the console less prominent (since folks often minimize it anyway).  In fact, I even have an issue to remove `console.log` from older samples; let's not add to the backlog of them.  https://github.com/OfficeDev/office-js-snippets/issues/28

samples/excel/45-named-item/create-and-use-named-item-for-range.yaml
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+        async function formatRangeByName() {
+            await Excel.run(async (context) => {
+
+                const sheet = context.workbook.worksheets.getItem("Sample");
+
+                const headerRowName = sheet.names.getItem("ExpensesHeader");
+                const headerRange = headerRowName.getRange();
+                headerRange.format.fill.color = "red";
```
Zlatkovsky: Would another hex color look better?  I assume that red looks quite contrastful...

samples/excel/45-named-item/create-and-use-named-item-for-range.yaml
```diff
@@ -0,0 +1,180 @@
...
+
+                let expensesTable = sheet.tables.add("A1:D1", true);
+                expensesTable.name = "ExpensesTable";
+
+                expensesTable.getHeaderRowRange().values = [["Date", "Merchant", "Category", "Amount"]];
+
+                let newData = transactions.map(item =>
+                    [item.date, item.merchant, item.category, item.amount]);
+
+                expensesTable.rows.add(null, newData);
```
Zlatkovsky: could you inline a comment for `null`, specifying what that parameter was?

    expensesTable.rows.add(null /*paramName*/, newData);

samples/excel/45-named-item/create-and-use-named-item-for-range.yaml
Rick-Kirkham: What is a "template string"?

samples/excel/45-named-item/create-and-use-named-item-for-range.yaml
Rick-Kirkham: The change from dark blue to red jumps out, so the user can easily see that the method worked.

samples/excel/45-named-item/create-and-use-named-item-for-range.yaml
Rick-Kirkham: I'll make these fixes, but FYI: **all** of your comments (including the blank lines) are a result of me just copying code from other samples. I presume that I won't be the only contributor who does that. People will take existing, accepted code as authoritative.

samples/excel/45-named-item/create-and-use-named-item-for-range.yaml
Rick-Kirkham: The JSON.stringify is needed or a complex object is what gets logged.

samples/excel/30-range/test-for-used-range.yaml
```diff
@@ -0,0 +1,125 @@
...
+        $("#setup").click(() => tryCatch(setup));
+        $("#try-create-chart-from-empty-table").click(() => tryCatch(tryCreateChartFromEmptyTable));
+
+        async function tryCreateChartFromEmptyTable() {
+            await Excel.run(async (context) => {
+
+                const sheet = context.workbook.worksheets.getItem("Sample");    
+                const salesTable = sheet.tables.getItem("SalesTable");
+                const dataRange = salesTable.getDataBodyRange();
+                const usedDataRange = dataRange.getUsedRangeOrNullObject(true /* only cells with values count as used */);
```
Zlatkovsky: I like the inline comment, but I meant something more like:

    const usedDataRange = dataRange.getUsedRangeOrNullObject(true /*valuesOnly*/);

And then if you want to add the comment of "only cells with values count as used", add it above or below.

samples/excel/30-range/test-for-used-range.yaml
```diff
@@ -0,0 +1,125 @@
...
+
+                //Must sync before reading value returned from *OrNullObject method/property.
+                await context.sync();
+
+                if (!usedDataRange.isNullObject) {
```
Zlatkovsky: please reverse `if` logic.  Shouldn't have a negative if followed by an `else`.

samples/excel/30-range/test-for-used-range.yaml
```diff
@@ -0,0 +1,125 @@
...
+
+                await context.sync();
+            });
+        }
+
+        async function setup() {
+            await Excel.run(async (context) => {
+                await OfficeHelpers.ExcelUtilities.forceCreateSheet(context.workbook, "Sample");
+                const sheet = context.workbook.worksheets.getItem("Sample");
+                let salesTable = sheet.tables.add('B2:F2', true);
```
Zlatkovsky: Nit: double-quotes

Zlatkovsky: Inline comment with parameter name for `true`

samples/excel/30-range/test-for-used-range.yaml
```diff
@@ -0,0 +1,125 @@
...
+    ExcelApi: 1.4
+script:
+    content: |
+        $("#setup").click(() => tryCatch(setup));
+        $("#try-create-chart-from-empty-table").click(() => tryCatch(tryCreateChartFromEmptyTable));
+
+        async function tryCreateChartFromEmptyTable() {
+            await Excel.run(async (context) => {
+
+                const sheet = context.workbook.worksheets.getItem("Sample");
```
Zlatkovsky: If you find the "const"-s with relation to variables odd (https://github.com/OfficeDev/office-js-snippets/issues/29), just replace them with "let".  I honestly think we over-pivoted on the "const" thing.  For my book, I've gone back and forth on this, but I use "let" most of the time because I think "const" is just confusing relative to what it means in other programming languages.

We can discuss real quick in person if you want.  Or just do what feels right to you for now, and we can then go back and take a look at what looks better (or send out a survey to MVPs and find out what they think)

samples/excel/30-range/test-for-used-range.yaml
```diff
@@ -0,0 +1,125 @@
...
+                if (!usedDataRange.isNullObject) {
+                    const chart = sheet.charts.add("ColumnClustered", dataRange, "auto");
+                    chart.setPosition("A15", "F30");
+                    chart.title.text = "Quarterly sales chart";
+                    chart.legend.position = "right"
+                    chart.legend.format.fill.setSolidColor("white");
+                    chart.dataLabels.format.font.size = 15;
+                    chart.dataLabels.format.font.color = "black";
+                } else {            
+                    OfficeHelpers.UI.notify("To create a meaningful chart, add names to the Product column and numbers to some of the other cells. Then press 'Try to create chart' again.");
```
Zlatkovsky: Notify has a method that takes in a title and a string.  I recommend you use that one.  It just looks nicer...

samples/excel/30-range/test-for-used-range.yaml
```diff
@@ -0,0 +1,125 @@
...
+                const usedDataRange = dataRange.getUsedRangeOrNullObject(true /* only cells with values count as used */);
+
+                //Must sync before reading value returned from *OrNullObject method/property.
+                await context.sync();
+
+                if (!usedDataRange.isNullObject) {
+                    const chart = sheet.charts.add("ColumnClustered", dataRange, "auto");
+                    chart.setPosition("A15", "F30");
+                    chart.title.text = "Quarterly sales chart";
+                    chart.legend.position = "right"
```
Zlatkovsky: missing semicolon

samples/excel/30-range/test-for-used-range.yaml
```diff
@@ -0,0 +1,125 @@
...
+                const sheet = context.workbook.worksheets.getItem("Sample");    
+                const salesTable = sheet.tables.getItem("SalesTable");
+                const dataRange = salesTable.getDataBodyRange();
+                const usedDataRange = dataRange.getUsedRangeOrNullObject(true /* only cells with values count as used */);
+
+                //Must sync before reading value returned from *OrNullObject method/property.
+                await context.sync();
+
+                if (!usedDataRange.isNullObject) {
+                    const chart = sheet.charts.add("ColumnClustered", dataRange, "auto");
```
Zlatkovsky: "auto"?  I would be more explicit and do by columns (esp since, IIRC, Office Online doesn't respect "auto" and just defaults to columns.)

Zlatkovsky: Also, for IntelliSense's sake (and people being able to play around and change out types), please use the enum-esque constants (`Excel.ChartType.ColumnsClustered`) rather than strings.

samples/excel/30-range/test-for-used-range.yaml
```diff
@@ -0,0 +1,125 @@
...
+                }
+
+                await context.sync();
+            });
+        }
+
+        async function setup() {
+            await Excel.run(async (context) => {
+                await OfficeHelpers.ExcelUtilities.forceCreateSheet(context.workbook, "Sample");
+                const sheet = context.workbook.worksheets.getItem("Sample");
```
Zlatkovsky: Why are you re-fetching the sheet?  You just created it using `forceCreateSheet`, set it to a variable and use it.  Same in other samples, if you used a similar pattern there.

samples/excel/45-named-item/create-and-use-named-item-for-range.yaml
```diff
@@ -0,0 +1,171 @@
...
+                const sheet = context.workbook.worksheets.getItem("Sample");
+                const headerRange = sheet.getRange("A1:E1");
+                sheet.names.add("ExpensesHeader", headerRange);
+                const namedItems = sheet.names.load("name, type");
+
+                await context.sync();
+
+                let namedItemsList = `This workbook contains ${namedItems.items.length} named item(s):`;
+                for (let i = 0; i < namedItems.items.length; i++) {
+                    namedItemsList += JSON.stringify(namedItems.items[i]);
```
Zlatkovsky: Still don't understand why you're using stringify.

samples/excel/45-named-item/create-and-use-named-item-for-range.yaml
```diff
@@ -0,0 +1,171 @@
...
+            await Excel.run(async (context) => {
+                const sheet = context.workbook.worksheets.getItem("Sample");
+                const headerRange = sheet.getRange("A1:E1");
+                sheet.names.add("ExpensesHeader", headerRange);
+                const namedItems = sheet.names.load("name, type");
+
+                await context.sync();
+
+                let namedItemsList = `This workbook contains ${namedItems.items.length} named item(s):`;
+                for (let i = 0; i < namedItems.items.length; i++) {
```
Zlatkovsky: This could be much neater with a `.map(item => item.name).join(", ");` instead of the for-loop.

Zlatkovsky: In which case the whole thing can become a single statement.

    utility.notify(<title>,
        `This workbook ... named item(s): ` +
        namedItems.items.map(...).join(", ")

samples/excel/45-named-item/create-and-use-named-item-for-range.yaml
```diff
@@ -0,0 +1,171 @@
...
+
+        async function setup() {
+            await Excel.run(async (context) => {
+                const sheet = await OfficeHelpers.ExcelUtilities.forceCreateSheet(context.workbook, "Sample");
+                const expensesTable = sheet.tables.add("A1:D1", true);
+                expensesTable.name = "ExpensesTable";
+                expensesTable.getHeaderRowRange().values = [["Date", "Merchant", "Category", "Amount"]];
+                const newData = transactions.map(item =>
+                    [item.date, item.merchant, item.category, item.amount]);
+                expensesTable.rows.add(null, // Row position (null = at the bottom)
```
Zlatkovsky: I would inline it using `/* */` syntax, and put the comma at the end of it.  and put in the parameter name as IntelliSense shows it first.  E.g.:

    null /*rowPosition (null = at the bottom)*/, newData

You can probably just move it to a new line just as I did above, and it should split reasonably nicely

samples/excel/45-named-item/create-and-use-named-item-for-range.yaml
```diff
@@ -0,0 +1,171 @@
...
+                                       newData);
+                sheet.getUsedRange().format.autofitColumns();
```
Zlatkovsky: You shouldn't be re-getching the same object (`getUsedRange()` twice, it's wasteful).  Always conserve your `get` methods.  Set to a usedRange variable instead.

samples/excel/45-named-item/create-and-use-named-item-for-range.yaml
```diff
@@ -0,0 +1,171 @@
...
+                await context.sync();
+            });
+        }
+
+        async function formatRangeByName() {
+            await Excel.run(async (context) => {
+                const sheet = context.workbook.worksheets.getItem("Sample");
+                const headerRowName = sheet.names.getItem("ExpensesHeader");
+                const headerRange = headerRowName.getRange();
+                headerRange.format.fill.color = "red";
```
Zlatkovsky: Don't know if you saw my earlier comment about the color "red".  I bet there is a more visually-appealing hex color.  And it's good to show hex being used, too...

samples/excel/30-range/test-for-used-range.yaml
Rick-Kirkham: I agree with this **""const" is just confusing relative to what it means in other programming languages."** But at this point we need to be consistent across all the snippets, so I'll stick with const for the most part.

PR #47, opened 2017-07-16
snippet to show how to use range intersections

Rick-Kirkham:


samples/excel/30-range/create-and-use-range-intersection.yaml
```diff
@@ -0,0 +1,154 @@
...
+                    reportMissingData("European");
+                } else {
+                    createContinentChart(sheet, "European", europeanContestRange, "A25", "F40");
+                }
+
+                await context.sync();
+            });
+        }
+
+        function createContinentChart(sheet: Excel.Worksheet, continent: string, contestRange: Excel.Range, startPosition: string, endPosition: string) {
```
Zlatkovsky: Nit: I would break up this line into two.  And also remove the newline below.

samples/excel/30-range/create-and-use-range-intersection.yaml
```diff
@@ -0,0 +1,154 @@
...
+                // Must sync before you can test the output of *OrNullObject 
+                // method/property.
+                await context.sync();
+
+                if (asianContestRange.isNullObject) {
+                    // See the declaration of this method for how to 
+                    // test this code path.
+                    reportMissingData("Asian");
+                } else {
+                    createContinentChart(sheet, "Asian", asianContestRange, "A9", "F24");
```
Zlatkovsky: I'm never in love with random strings like "A9" or "F25".  I think that if you can do them as relative references (`europenSalesRange.getCell(0,0).getOffsetRange(4, 0)`, or whatever, I think it makes it much easier to later modify the code.

samples/excel/30-range/create-and-use-range-intersection.yaml
```diff
@@ -0,0 +1,154 @@
...
+                    createContinentChart(sheet, "European", europeanContestRange, "A25", "F40");
+                }
+
+                await context.sync();
+            });
+        }
+
+        function createContinentChart(sheet: Excel.Worksheet, continent: string, contestRange: Excel.Range, startPosition: string, endPosition: string) {
+
+            let chart = sheet.charts.add("ColumnClustered", contestRange, "columns");
```
Zlatkovsky: Please use our enums, for better intellisense if folks want to change this out.

samples/excel/30-range/create-and-use-range-intersection.yaml
```diff
@@ -0,0 +1,154 @@
...
+            chart.setPosition(startPosition, endPosition);
+            chart.title.text = `${continent} Current Quarter Sales Contest`;
+            chart.legend.position = "right"
```
Zlatkovsky: Missing semicolon, and please use the enum.

samples/excel/30-range/create-and-use-range-intersection.yaml
Rick-Kirkham: Script Lab itself breaks the line at a space (and even indents the continuation of the line). Precisely because it is so long, I think it helps readability to have that blank line after it.

samples/excel/30-range/create-and-use-range-intersection.yaml
Rick-Kirkham: I agree in principle that offsets should be used in a production add-in. But this is a good example of a best practice that can actually interfere with the educational goal of a snippet by adding noise that isn't really relevant to the APIs that the snippet is trying to illustrate.

samples/excel/30-range/create-and-use-range-intersection.yaml
Rick-Kirkham: I found the enum (with some effort), but I'm going to hold off on making this change because our reference topic and the intellisense both seem to work very hard to hide the fact that the enum could be used (and even that it exists). If I type this:
`sheet.charts.add(`
The intellisense tells me that the first param is a string. There's no hint that an enum could be used. The reference topic for ChartCollection.add is the same way. I think using the enum will just confuse the heck out of people.

samples/excel/30-range/create-and-use-range-intersection.yaml
Rick-Kirkham: I added the semi-colon. Please see my response to your earlier comment about the enum.

PR #58, opened 2017-07-28
Add private OfficeJS snippets

cshashwat:
OfficeJS now has private snippets, which are independent of the old playlists (now treated as 'public') and will not be included as defaults in applications such as Script Lab.

The directory structure remains the same: under `private-samples` choose the appropriate `host`, add a subdirectory denoting what type of snippet it is, and then add the snippet name. A sample Word snippet is included in this pull request.

### DIFF
--- a/config/build.ts
+++ b/config/build.ts
@@ -78,7 +78,7 @@ const defaultApiSets = {
 
 
 async function processSnippets(dir) {
-    await new Promise((resolve, reject) => {
+    return new Promise((resolve, reject) => {
         banner('Loading & processing snippets');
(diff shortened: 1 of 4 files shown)


config/build.ts
```diff
@@ -40,31 +40,50 @@ const defaultApiSets = {
...
                 throw accumulatedErrors;
             }
         })
         .then(() => {
             banner('Done!', null, chalk.bold.green);
-            process.exit(0);
         })
         .catch(handleError);
+
+    processedSnippets.clear();
```
dsaunders45459: Please change this to a local variable instead of global, other than this change it looks good.

PR #60, opened 2017-08-03
Create -all playlists for hosts, containing public and private samples

cshashwat:


config/build.ts
```diff
@@ -56,12 +56,15 @@ const defaultApiSets = {
         })
         .catch(handleError);
 
-    let processedPrivateSnippets = new Dictionary<SnippetProcessedData>();
+    for (let processedSnippet of processedSnippets.values()) {
```
dsaunders45459: Why not set this in processSnippets when isPrivateSample is false

config/build.ts
```diff
@@ -157,7 +160,10 @@ async function processSnippets(dir, processedSnippets) {
                 order: (typeof (snippet as any).order === 'undefined') ? 100 /* nominally 100 */ : (snippet as any).order,
                 api_set: snippet.api_set
             };
-
+            if (isPrivateSample) {
```
dsaunders45459: Seems like we can set this directly in returned object:
return {
 id: snippet.id,
type: isPrivateSample? 'private' : 'public',
name: snippet.name

config/build.ts
```diff
@@ -91,7 +94,7 @@ async function processSnippets(dir, processedSnippets) {
 
     // Helpers:
 
-    async function processAndValidateSnippet(file: SnippetFileInput, dir: string): Promise<SnippetProcessedData> {
+    async function processAndValidateSnippet(file: SnippetFileInput, dir: string, isPrivateSample?: boolean): Promise<SnippetProcessedData> {
```
dsaunders45459: isPrivateSample shouldn't be optional, make it required.

config/build.ts
```diff
@@ -430,14 +436,16 @@ function checkSnippetsForUniqueIDs(processedSnippets) {
...
-    status.complete(true /*success*/, `Creating \'${dir}\' folder`);
+    if (!isPrivateSample) {
+        /* Creating playlists directory */
+        status.add(`Creating \'${dir}\' folder`);
+        await rmRf(dir);
+        await mkDir(dir);
+        status.complete(true /*success*/, `Creating \'${dir}\' folder`);
+    }
 
     const groups = groupBy(
```
dsaunders45459: The lines below this is probably why the order is random. You are deleting item.order and modifying the group name of processedSnippets. Instead of deleting in the dictionary directly, can groupBy return a new dictionary?

dsaunders45459: Sorry the code review does not seem to let me comment on existing lines. Below lines are causing the issue I believe.
        const groupNumberRegex = /^(\d+\s)?(\w.*)$/;

        items.forEach(item => {
            item.group = item.group.replace(groupNumberRegex, '$2');

            // Also remove "order", it's no longer needed (the snippets themselves are already in an ordered array in the YAML file)
            delete item.order;
        });

config/build.ts
cshashwat: Done.

config/build.ts
```diff
@@ -40,12 +40,12 @@ const defaultApiSets = {
 
 
 (async() => {
-    let processedPublicSnippets = new Dictionary<SnippetProcessedData>();
+    let processedSnippets = new Dictionary<SnippetProcessedData>();
     await Promise.resolve()
-        .then(() => processSnippets('samples', processedPublicSnippets))
+        .then(() => processSnippets('samples', processedSnippets, false))
```
Zlatkovsky: `true`/`false` parameters should always have an inline comment.  Ditto below

config/build.ts
```diff
@@ -56,12 +56,11 @@ const defaultApiSets = {
         })
         .catch(handleError);
 
-    let processedPrivateSnippets = new Dictionary<SnippetProcessedData>();
     await Promise.resolve()
```
Zlatkovsky: This looks like repeated code.  Why not process both directories at once?

config/build.ts
```diff
@@ -155,9 +154,9 @@ async function processSnippets(dir, processedSnippets) {
                 rawUrl: rawUrl,
                 group: startCase(file.group),
                 order: (typeof (snippet as any).order === 'undefined') ? 100 /* nominally 100 */ : (snippet as any).order,
```
Zlatkovsky: The "all" playlist needn't have an order

config/build.ts
```diff
@@ -91,7 +90,7 @@ async function processSnippets(dir, processedSnippets) {
 
     // Helpers:
 
-    async function processAndValidateSnippet(file: SnippetFileInput, dir: string): Promise<SnippetProcessedData> {
+    async function processAndValidateSnippet(file: SnippetFileInput, dir: string, isAllSample: boolean): Promise<SnippetProcessedData> {
```
Zlatkovsky: `isAllSample` = strange name.

Zlatkovsky: Why not make `dir` be an array (`dirs`), and process all dirs simultaneously?

playlists/excel.yaml
```diff
@@ -9,6 +9,7 @@
   group: Basics
   api_set:
     ExcelApi: 1.1
+  type: public
```
Zlatkovsky: Since this playlist is only going to have public snippets in it, maybe get rid of the `type` property before dumping to yaml?  Just for cleanliness sake...

Zlatkovsky: Actually, while you're at it, you could also get rid of `host`, since it's also redundant information (since each file is already grouped by host)

playlists/excel.yaml
```diff
@@ -9,6 +9,7 @@
...
 - id: excel-basic-api-call-es5
   name: Basic API call (JavaScript)
   fileName: basic-api-call-es5.yaml
```
Zlatkovsky: Are we ever using `filename` and `relativepath` in Script Lab?  If not, shall we get rid of it to save bandwidth?  Ditto for the `all` playlist.

playlists/excel-all.yaml
```diff
@@ -0,0 +1,582 @@
+- id: excel-basic-api-call
+  name: Basic API call
+  fileName: basic-api-call.yaml
+  relativePath: excel\01-basics\basic-api-call.yaml
+  description: Executes a basic Excel API call
```
Zlatkovsky: For the "all" playlist, I recommend we get rid of a bunch of the data, to make it lighter to download for the client.
Really, I think the only thing we care about is the `id` and `rawUrl`.  We don't care about the name or filename or relativePath or description or host or group or api_set or type...

private-samples/excel/01-basics/basic-modified-api-call.yaml
```diff
@@ -1,4 +1,4 @@
-id: excel-basic-api-call
```
Zlatkovsky: Do we actually need this one at all?  Or at least, can you name the file with a "TEMP" prefix, to make it clear that you're only doing this for testing, and that the file can be deleted as soon as we have something real to put in that folder instead?

config/build.ts
dsaunders45459: This one makes sense to call isPrivateSample/isPublicSample because it is just handling the samples under "samples" or "private-samples" directory and not processing combined playlist.

config/build.ts
```diff
@@ -430,21 +429,24 @@ function checkSnippetsForUniqueIDs(processedSnippets) {
     }
 }
 
-async function generatePlaylists(dir, processedSnippets: Dictionary<SnippetProcessedData>) {
+async function generatePlaylists(dir, processedSnippets: Dictionary<SnippetProcessedData>, isAllSample) {
```
dsaunders45459: Maybe a better name is isCombinedPlaylist

playlists/excel.yaml
dsaunders45459: It may not be a good idea to get rid of host, you would need to verify if Script Lab has a dependency on this property or you will break compatibility with this repo.

Zlatkovsky: Right, for any of those properties, you'll need to make sure they're not used by Script Lab.  But I don't think they are.  They are more probably a historical remnant.

cshashwat:
The "all" playlists are now pretty-printed JSON files under the 'view' directory, with each file containing a mapping from the snippet's id to its raw url.

config/build.ts
```diff
@@ -76,13 +61,14 @@ const defaultApiSets = {
 })();
 
 
-async function processSnippets(dir, processedSnippets) {
+async function processSnippets(processedSnippets) {
     return new Promise((resolve, reject) => {
         banner('Loading & processing snippets');
-        let files$ = getFiles(path.resolve(dir), path.resolve(dir));
+        let files$ = getFiles(path.resolve('private-samples'), path.resolve('private-samples'));
```
Zlatkovsky: I would put `private-samples` and `samples` as constants somewhere at the top of the file

config/build.ts
```diff
@@ -91,10 +77,11 @@ async function processSnippets(dir, processedSnippets) {
 
     // Helpers:
 
-    async function processAndValidateSnippet(file: SnippetFileInput, dir: string): Promise<SnippetProcessedData> {
+    async function processAndValidateSnippet(file: SnippetFileInput): Promise<SnippetProcessedData> {
         const messages: Array<string | Error> = [];
         try {
             status.add(`Processing ${file.relativePath}`);
+            let dir = file.type === 'private' ? 'private-samples' : 'samples';
```
Zlatkovsky: This feels like some weird reverse logic.  Shouldn't the file know its dir, rather than you deriving the path based on the file type.

config/helpers.ts
```diff
@@ -17,6 +17,7 @@ export interface SnippetFileInput {
     relativePath: string;
     host: string;
     group: string;
+    type: string;
```
Zlatkovsky: Why a string, rather than a boolean (`isPublic`)

config/build.ts
```diff
@@ -430,20 +417,27 @@ function checkSnippetsForUniqueIDs(processedSnippets) {
     }
 }
 
-async function generatePlaylists(dir, processedSnippets: Dictionary<SnippetProcessedData>) {
+async function generatePlaylists(processedSnippets: Dictionary<SnippetProcessedData>) {
     banner('Generating playlists');
 
+    let processedPublicSnippets = new Dictionary<SnippetProcessedData>();
+    for (let processedSnippet of processedSnippets.values()) {
+        if (processedSnippet.type === 'public') {
```
Zlatkovsky: I much prefer a boolean (or at minimum a strongly-typed string, but really just a boolean), for something like this.  That way, too, you can be guaranteed that the "else" statement will always be hit correctly, as opposed to being hit due to a misspelling.

config/build.ts
```diff
@@ -464,23 +458,55 @@ async function generatePlaylists(dir, processedSnippets: Dictionary<SnippetProce
...
+            delete item.type;
         });
 
-        let contents = jsyaml.safeDump(items, {
+        let contents = jsyaml.safeDump(modifiedItems, {
             skipInvalid: true /* skip "undefined" (e.g., for "order" on some of the snippets) */
         });
 
-        await writeFile(path.resolve(`${dir}/${host}.yaml`), contents);
+        /* Group private samples with public samples in one playlist */
```
Zlatkovsky: Whoa.  Shouldn't the yaml playlist only contain the public ones?  Maybe I'm just confused by the comment....

config/build.ts
```diff
@@ -464,23 +458,55 @@ async function generatePlaylists(dir, processedSnippets: Dictionary<SnippetProce
...
+    status.complete(true /*success*/, `Creating \'view\' folder`);
+
+    const allGroups = groupBy(
+        processedSnippets.values()
+            .filter((file) => !(file == null) && file.fileName !== 'default.yaml'),
+        'host');
+    let allPlaylistPromises = map(allGroups, async (items, host) => {
+        const creatingStatusText = `Creating ${host}.json`;
+        status.add(creatingStatusText);
+        items = sortBy(items, ['group', 'order', 'id']);
```
Zlatkovsky: `['group', 'order', 'id']` is used twice.  Should it be a constant at the top of the function, something like `const SORTING_CRITERIA = ['group', 'order', 'id'];`?

config/build.ts
```diff
@@ -464,23 +458,55 @@ async function generatePlaylists(dir, processedSnippets: Dictionary<SnippetProce
...
+
+        let hostMapping = {};
```
Zlatkovsky: Could you type it as a TypeScript dictionary?

    const hostMapping = {} as { [id: string]: string };

config/helpers.ts
```diff
@@ -201,6 +203,7 @@ export const getFileMetadata = (file: string, root: string) => {
 
     return Observable.of<SnippetFileInput>({
         relativePath: relativePath,
+        type: /private-samples/.test(file) ? 'private' : 'public',
```
Zlatkovsky: This is unsafe.  What if I called a file "make-private-samples-worksheet.yaml".  If you're using a regular expression, be sure to specify its start and end conditions.  Or just use something else, e.g., an API to get the parent of the file, or to at least only get the path portion of the file rather than the file name.

playlists/excel.yaml
```diff
@@ -14,7 +13,6 @@
...
   relativePath: excel\01-basics\basic-api-call-es5.yaml
```
Zlatkovsky: Is this "relativePath" property used anywhere (e.g., in Script Lab?).  I don't think so.  If so, let's remove it, and cut down on the bandwidth...

config/build.ts
```diff
@@ -464,23 +458,55 @@ async function generatePlaylists(dir, processedSnippets: Dictionary<SnippetProce
         */
         const groupNumberRegex = /^(\d+\s)?(\w.*)$/;
 
-        items.forEach(item => {
+        let modifiedItems = cloneDeep(items);
```
dsaunders45459: Instead of cloning and deleting items that we do not want, I think it may be cleaner to do a map on the array instead and select the items we want. Something like:
playlistItem = items.map({name, description, etc} => {
return {name, description, etc};
});

playlists/excel.yaml
cshashwat: Removed.

config/helpers.ts
cshashwat: Strengthened the regex and changed from testing `file` to testing `root`, which only contains the path-names.

config/helpers.ts
cshashwat: Changed to boolean throughout code.

config/build.ts
cshashwat: Done.

config/build.ts
cshashwat: Added.

config/build.ts
cshashwat: Comment should have been removed, thanks for catching it. Fixed.

config/build.ts
cshashwat: Has been addressed with `isPublic`.

config/build.ts
cshashwat: Added.

config/build.ts
cshashwat: Added.

PR #61, opened 2017-08-04
Merge master -> prod

Rick-Kirkham:


README.md
```diff
@@ -7,39 +7,77 @@ A collection of code snippets built with [Script Lab](github.com/OfficeDev/scrip
...
-    2. Right-click the file name under **Changes** and select **Open changes**. Both versions of the file will be open in the editor so you can see the pending changes relative to the staged version. You may find that the script already substituted in required fields like `id` or `api_set` with reasonable defaults. 
-12. Re-run `npm start`, and fix errors, until the build succeeds. 
-13. Submit to the repo, and create a merge request into master.
+ > Note: In a few days, the repo admins will merge your snippet into the `prod` branch. It will then appear in **Samples** area of Script Lab. (It is in the **My Snippets** area as soon as you create it.) 
+
+25. Optionally, you can delete the branch you created from your fork and/or your local clone.
+
+#### Known errors and fixes in the build tool
+- An error saying that `name` has upper-case letters or other disallowed characters is *not* referring to the `name` property in the file. It is referring to the file name itself. You'll also get this error, if the file extension is not `.yaml`.
```
Zlatkovsky: Do you want to just fix the tool? :-)  I can show you how, if you're interested.  Will be available in the afternoon...

config/build.ts
```diff
@@ -40,31 +39,50 @@ const defaultApiSets = {
 };
 
 
-(() => {
-    Promise.resolve()
-        .then(processSnippets)
+(async() => {
```
Zlatkovsky: Ricky, I am not 100% sure that this code is fully ready yet, I know that Shashwat is working on it.  You may be best off waiting till later today, or merging in (cherry-picking?) only the samples files but not the build tool.  Or at least, making sure that the playlist produced by the new changes to the tooling is no different than it was before.

playlists/excel.yaml
```diff
@@ -136,6 +158,19 @@
   group: Range
   api_set:
     ExcelApi: 1.1
+- id: excel-35-worksheet-reference-worksheets-by-relative-position-
```
Zlatkovsky: Are we generating such name?  with the folder name (`35-worksheet`) within it?  That seems odd...

Rick-Kirkham: Yes. I think that's expected behavior to help ensure unique ids.

config/build.ts
Rick-Kirkham: Hmmm. I already merged this to prod. We thought we had your permission to do that while you were gone. My new manager (Reeza) wanted me to do it right away, so I just did. It did not occur to me that code that isn't ready would be in master. I accepted that PR at Dan Saunders request.

Zlatkovsky: That's fine, you do have my permission :-).  As long as the playlist looks good (which actually I just validated, and it does), that's fine.  I just know that Shashwat is currently making some changes to the deploy script, so I wanted to make sure this wouldn't be impacted.  But it looks like it's not.  Don't worry about it.

Zlatkovsky: Just be sure to check in the "prod" script lab that your new samples are showing up, and at the same time that everything else looks ok (i.e., that the groups are in the right order, and that the snippets within are in the right order as well.  A 20-second smoke-test).

Rick-Kirkham: Right. I verified that it looks right in prod right after I pulled it to prod branch. It looks fine, I think.

PR #63, opened 2017-08-07
Add support for outlook

annich-MS:
A small number of snippets to start with. More to be added later


samples/outlook/01-compose-basics/get-item-subject.yaml
```diff
@@ -0,0 +1,48 @@
+id: outlook-01-compose-basics-get-item-subject
+name: Get Item Subject
+description: Gets the subject of a compose item
+host: OUTLOOK
+api_set: {}
+script:
+    content: |
+        $("#run").click(run);
+
```
Zlatkovsky: I would remove the extra newline, and only leave one.  Ditto for the next samples.

samples/outlook/01-compose-basics/get-item-subject.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+        function run() {
+            let item = Office.context.mailbox.item as Office.ItemCompose;
+            item.subject.getAsync((result: Office.AsyncResult) => {
+                if (result.status !== Office.AsyncResultStatus.Succeeded) {
+                    OfficeHelpers.UI.notify(`Action failed with message ${result.error.message}`);
+                } else {
```
Zlatkovsky: What about just `return`-in out within the `if` statement.  And then you don't need the `else` and the nesting here.  Ditto for the next samples.

samples/outlook/01-compose-basics/get-item-subject.yaml
```diff
@@ -0,0 +1,48 @@
...
+        <button id="run" class="ms-Button">
+            <span class="ms-Button-label">Run</span>
+        </button>
+    language: html
+style:
+    content: "/* Your style goes here */\r\n"
+    language: css
+libraries: |
+    # Office.js
+    https://appsforoffice.microsoft.com/lib/1/hosted/office.js
```
Zlatkovsky: Didn't you need the private URLs?

PR #64, opened 2017-08-09
Prevent build.ts from overwriting office.js in outlook snippets

annich-MS:
Noticed that npm start was overwriting the custom office.js

### DIFF
--- a/config/build.ts
+++ b/config/build.ts
@@ -237,7 +237,8 @@ async function processSnippets(processedSnippets) {
             throw new Error(`Cannot have more than one reference to Office.js or ${officeDTS}`);
         }
 
-        if (officeJsReferences[0] !== canonicalOfficeJsReference) {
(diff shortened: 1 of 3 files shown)


config/build.ts
```diff
@@ -237,7 +237,8 @@ async function processSnippets(processedSnippets) {
             throw new Error(`Cannot have more than one reference to Office.js or ${officeDTS}`);
         }
 
-        if (officeJsReferences[0] !== canonicalOfficeJsReference) {
+        // for now, outlook does not want to use cannonical Office.js due to a bug.
```
Zlatkovsky: Could you file the bug and put is number here?  And likewise, add a note in the bug that we need to both fix up the samples and remove the bit of code added by this commit, once the Outlook change percolates up.

BTW, haven't seen you submit anything internally for update the Outlook files on the CDN...

PR #67, opened 2017-08-21
Multiple property set snippet for Word

Rick-Kirkham:


playlists/word.yaml
```diff
@@ -16,6 +16,15 @@
   group: Basics
   api_set:
     WordApi: 1.1
+- id: word-01-basics-multiple-property-set-
```
Zlatkovsky: Ricky, could you please move this out of `Basics` and... not quite sure where :-)?  Maybe we need a "general patterns" group.  But Basics is definitely the wrong place, and causes clutter.

Ditto for the Excel one.

samples/word/01-basics/multiple-property-set.yaml
```diff
@@ -0,0 +1,123 @@
...
+        $("#copy-properties-from-paragraph").click(() => tryCatch(copyPropertiesFromParagraph));
+
+        async function setMultiplePropertiesWithObject() {
+            await Word.run(async (context) => {
+                const paragraph = context.document.body.paragraphs.getFirst();
+                paragraph.set({
+                    leftIndent: 30,
+                    font: {
+                        bold: true,
+                        color: 'red'
```
Zlatkovsky: For consistency, double-quotes, not single quotes

PR #71, opened 2017-08-24
New snippet for get/set worksheet tab color

kbrandl:
Added new snippet that shows how to get/set worksheet tab color; also renamed the gridlines.yaml snippet file to "worksheet-gridlines".


samples/excel/89-preview-apis/worksheet-tab-color.yaml
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        async function getTabColor() {
+            await Excel.run(async (context) => {
+                const activeSheet = context.workbook.worksheets.getActiveWorksheet();
+                activeSheet.load("name, tabColor");
+
+                await context.sync();
+
+                let tabColor = activeSheet.tabColor !== "" ? activeSheet.tabColor : "the default color";
```
Zlatkovsky: Please reverse the !==.  A negative with an else is a double negative.  So do a comparison for `=== ""` instead.

PR #72, opened 2017-08-24
New snippet for  get/set text orientation within a range

kbrandl:
Add new snippet that shows how to get and set text orientation within a range; edit worksheet-gridlines snippet to remove log statements.

### DIFF
--- a/samples/excel/89-preview-apis/range-text-orientation.yaml
+++ b/samples/excel/89-preview-apis/range-text-orientation.yaml
@@ -1,0 +1,120 @@
+id: excel-89-preview-apis-range-text-orientation-
+name: Range text orientation
+description: Set and get the text orientation within a range
+author: OfficeDev
+host: EXCEL
+api_set:
+    ExcelApi: 1.7
+script:
+    content: |
+        $("#setup").click(() => tryCatch(setup));
+        $("#set-text-orientation").click(() => tryCatch(setTextOrientation));
+        $("#get-text-orientation").click(() => tryCatch(getTextOrientation));
+
+        async function setTextOrientation() {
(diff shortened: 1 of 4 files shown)


samples/excel/89-preview-apis/range-text-orientation.yaml
```diff
@@ -0,0 +1,120 @@
...
+
+        async function getTextOrientation() {
+            await Excel.run(async (context) => {
+                const sheet = context.workbook.worksheets.getItem("Sample");
+                const range = sheet.getRange("B2:E2");
+                range.load("address, format/textOrientation");
+
+                await context.sync();
+
+                let textOrientation = range.format.textOrientation !== null ? range.format.textOrientation : "null (which indicates that the text orientation for cells within the specified range is not uniform)";
```
Zlatkovsky: Please reverse the "!==" and also multi-line (since it's a long text)

    let textOrientation = range.format.textOrientation === null ?
        range.format.textOrientation : 
        "null (which indicates that ...)"

samples/excel/89-preview-apis/range-text-orientation.yaml
```diff
@@ -0,0 +1,120 @@
...
+
+                console.log(`The text orientation within the range "${range.address}" is ${textOrientation}.`);
```
Zlatkovsky: `console.log` or `notify`?
I still maintain that if it's only a single notification for the entire run of the user action, I would do `notify`.  If it's multiple, and it's worthwhile to see the history, *then* I would use `console.log` (but potentially still have a single `notify` at the end to inform someone to look at the log)

PR #73, opened 2017-08-25
New snippet for freeze/unfreeze panes in a worksheet

kbrandl:
Add new snippet that shows how to freeze/unfreeze columns and rows and manage frozen panes in a worksheet.


samples/excel/89-preview-apis/worksheet-freeze-panes.yaml
```diff
@@ -0,0 +1,211 @@
...
+        $("#unfreeze-all-columns").click(() => tryCatch(unfreezeAllColumns));
+        $("#freeze-rows").click(() => tryCatch(freezeRows));
+        $("#unfreeze-all-rows").click(() => tryCatch(unfreezeAllRows));
+        $("#freeze-at").click(() => tryCatch(freezeAt));
+        $("#get-location").click(() => tryCatch(getLocation));
+        $("#unfreeze-all-panes").click(() => tryCatch(unfreezeAllPanes));
+
+        async function freezeColumns() {
+            await Excel.run(async (context) => {
+                const activeSheet = context.workbook.worksheets.getActiveWorksheet();
```
Zlatkovsky: Active sheet or get sheet by name?

kbrandl: Good call -- I probably should get sheet by name, since the "setup" function explicitly creates a "Sample" sheet. I'll make this change in my other snippets for this as well.

PR #94, opened 2017-09-26
Fix some issues around hyphenation of IDs, and numerics in the IDs.

Zlatkovsky:
Also strip out "preview-apis-" from name, since those APIs, sooner or later, will not be preview.

The only real change is in `config/build.ts`, after that I simply ran "npm start".  I think the result looks correct, but want someone else to look at it just in case.


config/build.ts
```diff
@@ -393,22 +393,24 @@ async function processSnippets(processedSnippets) {
...
-                .replace(/^_+/, '') /* trim any hyphens before */
-                .replace(/_+$/, '') /* and trim any at the end, as well */;
-
-            messages.push('Snippet ID may not be empty or be a machine-generated ID.');
-            messages.push(`... replacing with an ID based on name: "${snippet.id}"`);
-        }
+                .replace(/^-+/, '') /* trim any hyphens before */
+                .replace(/-+$/, '') /* and trim any at the end, as well */
+                .replace(/-(\d+-)(.*)/, '-$2') /* remove any numeric prefixes like "word\01-basics\foo", replacing with "word\basics\foo" */
+                .replace('-preview-apis-', '-');
```
kbrandl: Is it really necessary to remove "-preview-apis-" from the snippet ID?  

When an API moves from "Preview" to "Prod", we'll be moving the snippet yaml file out of the "Preview APIs" folder and into the appropriate object-specific folder ("Worksheet", "Range", etc.) -- at which point (if I'm understanding this function correctly), the snippet ID will be regenerated to reflect the new location of the snippet (i.e., the ID will no longer contain "-preview-apis-").

PR #95, opened 2017-09-28
Update range-hyperlink snippet

kbrandl:
(Updated snippet per Sudhi's feedback.)

### DIFF
--- a/samples/excel/89-preview-apis/range-hyperlink.yaml
+++ b/samples/excel/89-preview-apis/range-hyperlink.yaml
@@ -10,119 +10,265 @@ script:
     content: |
         $("#setup").click(() => tryCatch(setup));
 
-        $("#create-url-hyperlink").click(() => tryCatch(createUrlHyperlink));
-        $("#create-document-hyperlink").click(() => tryCatch(createDocumentHyperlink));
-        $("#update-hyperlink").click(() => tryCatch(updateHyperlink));
-        $("#clear-hyperlink").click(() => tryCatch(clearHyperlink));
+        $("#create-url-hyperlinks").click(() => tryCatch(createUrlHyperlinks));
+        $("#create-document-hyperlinks").click(() => tryCatch(createDocumentHyperlinks));
+        $("#update-hyperlinks").click(() => tryCatch(updateHyperlinks));
+        $("#clear-hyperlinks").click(() => tryCatch(clearHyperlinks));
 
-        async function createUrlHyperlink() {
+        async function createUrlHyperlinks() {
             await Excel.run(async (context) => {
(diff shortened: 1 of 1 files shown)


kbrandl:
@Rick-Kirkham -- I've updated this snippet per Sudhi's feedback. Can you please review (whenever you have time)?  Thanks!

samples/excel/89-preview-apis/range-hyperlink.yaml
```diff
@@ -10,121 +10,267 @@ script:
...
-                range.hyperlink = hyperlink;
 
                 await context.sync();
+
+                OfficeHelpers.UI.notify("Created a hyperlink to a location in this workbook for each of the products in the second table.");
             });
         }
 
-        async function updateHyperlink() {
+        async function updateHyperlinks() {
```
Rick-Kirkham: Not mandatory, but IMHO, this is essentially identical to the create* method. I don't think we need both. Can we just have a comment above the create* method that says something like "use this to create or update a hyperlink"?

samples/excel/89-preview-apis/range-hyperlink.yaml
```diff
@@ -10,121 +10,267 @@ script:
...
+                createMelonData(melonSheet, context);
 
-                sheet.activate();
+                ordersSheet.activate();
 
                 await context.sync();
             });
         }
 
+        async function createOrdersData(ordersSheet: Excel.Worksheet, context: Office.Context) {
```
Rick-Kirkham: All of the create*Data() methods have the `async `keyword, but they don't contain anything awaitable. Apparently, the compiler let's you get away with this, but I think it might confuse developers. Unless there's some reason I'm not aware of for having the `async `keyword on these methods, can we remove it?

samples/excel/89-preview-apis/range-hyperlink.yaml
kbrandl: You're right, it just so happens that the process for creating a range hyperlink and updating a range hyperlink are identical.  Creating separate functions (and having separate buttons) for create and update was done per Sudhi's request, so probably best to leave it this way.

samples/excel/89-preview-apis/range-hyperlink.yaml
kbrandl: Good catch!  Originally, I was doing sheet creation in each of these functions (so I needed to pass-in a **context** object and include the word **async** in the signature)....but since I moved the sheet creation statements into the setup() function instead, the *Data() functions no longer need to be **async** or to accept a **context** object as an input parameter. I'll make these changes...thanks!

PR #96, opened 2017-09-29
Shows how to get and set document properties

siewmoi:


samples/excel/89-preview-apis/document-properties.yaml
```diff
@@ -0,0 +1,118 @@
...
+                console.log("Last author  : " + docProperties.lastAuthor);
+                console.log("Revision number : " + docProperties.revisionNumber);
+              
+                OfficeHelpers.UI.notify("Workbook creation date : " + docProperties.creationDate.toDateString());            
+            });
+        }
+
+        async function setGetDocProperties() {
+            await Excel.run(async (context) => {
+                let titleValue = "Excel document properties API";
```
Rick-Kirkham: We generally use `const` instead if `let` in our snippets for variables that won't get written to again. But that's somewhat controversial as guidance, so you can leave it if you prefer `let`.

samples/excel/89-preview-apis/document-properties.yaml
```diff
@@ -0,0 +1,118 @@
...
+                docProperties.comments = commentsValue;
+                docProperties.category = categoryValue;
+                docProperties.manager = managerValue;
+                docProperties.company = companyValue;
+
+                await context.sync();
+                
+                // Get the docoment properties.
+                // To learn how to view document properties in the Ui, 
+                // see https://support.office.com/en-us/article/View-or-change-the-properties-for-an-Office-file-21D604C2-481E-4379-8E54-1DD4622C6B75
```
Rick-Kirkham: Good comment.

samples/excel/89-preview-apis/document-properties.yaml
```diff
@@ -0,0 +1,118 @@
...
+                let subjectValue = "Set and get document properties";
+                let keywordsValue = "Set and get operations";
+                let commentsValue = "This is an Excel document properties API code sample";
+                let categoryValue = "Office Add-ins";
+                let managerValue = "John";
+                let companyValue = "Microsoft";
+
+                let docProperties = context.workbook.properties;
+
+                // Set the docoment properties.
```
Rick-Kirkham: The code here is just creating and writing to properties of a local variable. It's not writing that variable back to `context.workbook.properties`, so it's not really showing how to set the document properties.

samples/excel/89-preview-apis/document-properties.yaml
```diff
@@ -0,0 +1,118 @@
...
+                docProperties.category = categoryValue;
+                docProperties.manager = managerValue;
+                docProperties.company = companyValue;
+
+                await context.sync();
+                
+                // Get the docoment properties.
+                // To learn how to view document properties in the Ui, 
+                // see https://support.office.com/en-us/article/View-or-change-the-properties-for-an-Office-file-21D604C2-481E-4379-8E54-1DD4622C6B75
+                console.log("Title :" + docProperties.title);
```
Rick-Kirkham: The code here is just reading back the local variable. It's not reading the actual `context.workbook.properties`.

I think what you need is a method that sets `context.workbook.properties`. Then have an instruction in the taskpane that tells the user to click the **Get document properties** button a second time.

PR #100, opened 2017-10-10
Chart axis

siewmoi:


samples/excel/89-preview-apis/chart-axis.yaml
```diff
@@ -6,7 +6,7 @@ host: EXCEL
...
...t: Excel.RequestContext) {\n    const sheet = context.workbook.worksheets.getItem(\"Sample\");\n    const salesTable = sheet.tables.getItem(\"SalesTable\");\n\n    const dataRange = salesTable.getRange();\n\n    let chart = sheet.charts.add(\"ColumnClustered\", dataRange, Excel.ChartSeriesBy.columns);\n\n    chart.setPosition(\"A15\", \"F30\");\n    chart.title.text = \"Quarterly sales chart\";\n    chart.legend.position = \"right\"\n    chart.legend.format.fill.setSolidColor(\"white\");\n    chart.dataLabels.format.font.size = 15;\n    chart.dataLabels.format.font.color = \"black\";\n    let points = chart.series.getItemAt(0).points;\n}\n\nasync function addVerticalAxisLabel(context: Excel.RequestContext) {\n\n        const sheet = context.workbook.worksheets.getItem(\"Sample\");\n\n        let chart = sheet.charts.getItemAt(0);\n        let axis = chart.axes.valueAxis;\n        axis.displayUnit = \"Thousands\";\n}\n\n/** Default helper for invoking an action and handling errors. */\nasync function tryCatch(callback) {\n    try {\n        await callback();\n    }\n    catch (error) {\n        OfficeHelpers.UI.notify(error);\n        OfficeHelpers.Utilities.log(error);\n    }\n}\n"
```
Rick-Kirkham: In the changeAxisUnit function, I think you should just remove the call to OfficeHelpers.UI.notify, for two reasons: (1) The term "Hundreds" appears in the chart and that's all the proof you need that the method worked. (2) Because you don't load displayUnit before you read it in the call to OfficeHelpers.UI.notify, all you are really reading is the proxy object, so this wouldn't prove that the code changed the chart anyway.

Rick-Kirkham: Remove the "async" keyword from the createChart and addVerticalAxisLabel functions. They don't have anything awaitable in them.

PR #106, opened 2017-10-23
Shows how to add, apply, get and delete styles.

siewmoi:


samples/excel/89-preview-apis/style.yaml
```diff
@@ -0,0 +1,241 @@
...
+                let styles = context.workbook.styles;
+
+                // Add a new style to the style collection.
+                // Styles is in the Home tab ribbon.
+                styles.add("Diagonal Orientation Style");
+                styles.getItem("Diagonal Orientation Style").orientation = 38;
+                styles.getItem("Diagonal Orientation Style").addIndent = true;
+               
+                let newStyle = styles.getItem("Diagonal Orientation Style");
+                context.load(newStyle);
```
Rick-Kirkham: The context.load(newStyle); is not needed because you are not reading the new style's properties, only writing to them. The function works fine if I comment the line out.

samples/excel/89-preview-apis/style.yaml
```diff
@@ -0,0 +1,241 @@
...
+        $("#delete-new-style").click(() => tryCatch(deleteNewStyle));
+
+        async function addNewStyle() {
+            await Excel.run(async (context) => {
+                let styles = context.workbook.styles;
+
+                // Add a new style to the style collection.
+                // Styles is in the Home tab ribbon.
+                styles.add("Diagonal Orientation Style");
+                styles.getItem("Diagonal Orientation Style").orientation = 38;
```
Rick-Kirkham: Why do you set the orientation and addIndent up here instead of down where you set the other 3 properties? If there's a good reason, please use a code comment to explain. I find that the function works fine if I comment out these 2 lines and add the following 2 lines down where the other properties are set:
        newStyle.orientation = 38;
        newStyle.addIndent = true;

samples/excel/89-preview-apis/style.yaml
```diff
@@ -0,0 +1,241 @@
...
+            await Excel.run(async (context) => {
+                let worksheet = context.workbook.worksheets.getItem("Sample");
+                context.load(worksheet);
+
+                // Apply new style.
+                worksheet.getRange("A1:E1").style = ("Diagonal Orientation Style");
+                worksheet.getRange("A1:E1").format.verticalAlignment = "Justify";
+
+                let range = worksheet.getRange("A1:E1");
+                context.load(range);
```
Rick-Kirkham: These 2 loads don't serve any purpose because you don't read any range properties after the sync. The function runs fine if I comment them out.

samples/excel/89-preview-apis/style.yaml
```diff
@@ -0,0 +1,241 @@
...
+        }
+
+        async function applyBuiltInStyle() {
+            await Excel.run(async (context) => {
+                let worksheet = context.workbook.worksheets.getItem("Sample");
+                context.load(worksheet);
+
+                // Apply built-in style. 
+                // Styles are in the Home tab ribbon.
+                worksheet.getRange("A1:E1").style = Excel.BuiltInStyle.neutral;
```
Rick-Kirkham: The code gets a range object for the exact same range 3 times. You can get it once and assign to a variable. Then set the style and verticalAlignment properties using the variable.

samples/excel/89-preview-apis/style.yaml
```diff
@@ -0,0 +1,241 @@
...
+                context.load(range.format);
+
+                await context.sync();
+            });
+        }
+
+        async function getStyleFontProperties() {
+            await Excel.run(async (context) => {
+                let style = context.workbook.styles.getItem("Normal");
+                context.load(style);
```
Rick-Kirkham: You don't need to load the style object when you are loading specific properties of it anyway. So this line isn't needed.

samples/excel/89-preview-apis/style.yaml
```diff
@@ -0,0 +1,241 @@
...
+                style.font.load("bold, color, italic, name, size");
```
Rick-Kirkham: I think it would be helpful to developers to know the "forward slash" syntax that they can use to load sub-properties. You can replace all 3 of your loads with just this line:
style.load('font/bold, font/color, font/italic, font/name, font/size, fill/color');

samples/excel/89-preview-apis/style.yaml
```diff
@@ -0,0 +1,241 @@
...
+                console.log("Name: " + style.font.name);
+                console.log("Size: " + style.font.size);
+                console.log("Fill color: " + style.fill.color);
+            });
+        }
+
+        async function getStyleAlignmentProperties() {
+            await Excel.run(async (context) => {
+                let style = context.workbook.styles.getItem("Diagonal Orientation Style");            
+                context.load(style);
```
Rick-Kirkham: Note that context.load(someobject) is just as much an empty load as someobject.load(); 
It is a good practice to avoid empty loads. It would be better to have:
style.load('orientation, horizontalAlignment, ... and so on );

samples/excel/89-preview-apis/style.yaml
```diff
@@ -0,0 +1,241 @@
...
+                console.log("Include protection: " + style.includeProtection);
+                console.log("Shrink to fit: " + style.shrinkToFit);
+                console.log("Style locked: " + style.locked);
+            });
+        }
+
+        async function deleteNewStyle() {
+            await Excel.run(async (context) => {
+                let style = context.workbook.styles.getItem("Diagonal Orientation Style");
+                context.load(style);
```
Rick-Kirkham: You don't need this load; because deleting an object doesn't require reading it back to the task pane.

PR #128, opened 2018-01-30
Add eventArgs example to Excel data changed sample

deinspanjer:
The docs don't do a great job explaining how to use this to full effect, and because the current sample doesn't even include the eventArgs argument, a user doesn't get a sense of how to interact with it.


kbrandl:
@deinspanjer thanks for the contribution!  I've made one minor update and will publish this snippet update momentarily.

PR #133, opened 2018-02-06
tweak workbook-get-active-cell sample

ElizabethSamuel-MSFT:


samples/excel/89-preview-apis/workbook-get-active-cell.yaml
```diff
@@ -34,9 +34,16 @@ script:
     language: typescript
 template:
     content: |
-        <button id="get-active-cell" class="ms-Button">
-            <span class="ms-Button-label">Get active cell</span>
-        </button>
+        <section class="ms-font-m">
+            <p>This sample shows how to get the active cell of the entire workbook.</p>
```
Rick-Kirkham: I think the word "entire" is confusing here and not needed.

samples/excel/89-preview-apis/workbook-get-active-cell.yaml
```diff
@@ -11,8 +11,8 @@ script:
         async function run() {
             await Excel.run(async (context) => {
 
-                const myWorkbook = context.workbook;
-                const activeCell = myWorkbook.getActiveCell();
+                let myWorkbook = context.workbook;
```
Rick-Kirkham: I think `const `is TypeScript best practice when the value is not going to change. We haven't been consistent on this in the past, but I prefer `const`.

PR #134, opened 2018-02-07
tweak worksheet-copy sample

ElizabethSamuel-MSFT:


samples/excel/89-preview-apis/worksheet-copy.yaml
```diff
@@ -12,16 +12,16 @@ script:
         async function run() {
             await Excel.run(async (context) => {
 
-                const myWorkbook = context.workbook;
-                const sampleSheet = myWorkbook.worksheets.getActiveWorksheet();
-                const copiedSheet = sampleSheet.copy("end")
+                let myWorkbook = context.workbook;
```
Rick-Kirkham: I prefer const which is a best practice if the value isn't going to change. But I'll approve anyway. This is just a recommendation.

PR #140, opened 2018-02-14
Create additional chart types

Rick-Kirkham:


siewmoi:
Ricky -- I've put Juan as reviewer because he requested specific scenarios.

PR #143, opened 2018-02-15
shows how to protect workbook structure and data in worksheet

ElizabethSamuel-MSFT:


samples/excel/89-preview-apis/protect-data-in-worksheet-and-workbook-structure.yaml
```diff
@@ -0,0 +1,258 @@
...
+            
+            let message = "not be protected";
+            if (expectProtection) {
+                message = "be protected";
+            }
+
+            return message;
+        }
+
+        async function passwordHelper(expectPassword: boolean, password?: string) {
```
Rick-Kirkham: You have the `async `keyword on this passwordHelper definition, but it doesn't look like it contains anything awaitable. I don't think the `async `is needed.

samples/excel/89-preview-apis/protect-data-in-worksheet-and-workbook-structure.yaml
```diff
@@ -0,0 +1,258 @@
...
+                    workbook.protection.protect(password);
+                } else if (!expectProtection) {
+                    workbook.protection.unprotect(password);
+                }
+
+                workbook.load("protection/protected");
+
+                await context.sync();
+
+                console.log("Expected: This workbook should " + await messageHelper(expectProtection) + ". Protected? " + workbook.protection.protected);
```
Rick-Kirkham: I don't see the purpose in making messageHelper asynchronous. It's just string concatenation.

samples/excel/89-preview-apis/protect-data-in-worksheet-and-workbook-structure.yaml
```diff
@@ -0,0 +1,258 @@
...
+
+        async function passwordUnprotectWorkbookStructure() {
+            await Excel.run(async (context) => {
+                await workbookHelper(context, false, true, await passwordHandler());
+            });
+        }
+
+        async function worksheetHelper(context: Excel.RequestContext, expectProtection: boolean, expectPassword: boolean, password?: string) {
+
+            await passwordHelper(expectPassword, password);
```
Rick-Kirkham: If you removed `async` from the definition of passwordHelper, you would not need the `await` here. Does passwordHelper really need to be defined as async?

samples/excel/89-preview-apis/protect-data-in-worksheet-and-workbook-structure.yaml
```diff
@@ -0,0 +1,258 @@
...
+        $("#protect-workbook-structure").click(() => tryCatch(protectWorkbookStructure));
+        $("#unprotect-workbook-structure").click(() => tryCatch(unprotectWorkbookStructure));
+        $("#password-protect-data-in-worksheet").click(() => tryCatch(passwordProtectDataInWorksheet));
+        $("#password-unprotect-data-in-worksheet").click(() => tryCatch(passwordUnprotectDataInWorksheet));
+        $("#password-protect-workbook-structure").click(() => tryCatch(passwordProtectWorkbookStructure));
+        $("#password-unprotect-workbook-structure").click(() => tryCatch(passwordUnprotectWorkbookStructure));
+
+        async function protectDataInWorksheet() {
+            await Excel.run(async (context) => {
+                await worksheetHelper(context, true, false);
```
Rick-Kirkham: My main concern is that you are nesting methods that call Excel.run inside methods that call Excel.run. It looks like the only purpose of the outer Excel.run is to get the context object. This seems like an unnecessary complication. If you keep the nested structure, is there a way you can rewrite the inner method so that it does not call Excel.run itself?

samples/excel/89-preview-apis/protect-data-in-worksheet-and-workbook-structure.yaml
```diff
@@ -0,0 +1,258 @@
...
+            });
+        }
+
+        async function unprotectDataInWorksheet() {
+            await Excel.run(async (context) => {
+                await worksheetHelper(context, false, false);
```
Rick-Kirkham: One problem with "Helper" methods is that they are not self-documenting. (Help what?) And they often violate the principle that any method should do just one thing. Your helpers do multiple things, including opposite things; e.g. they protect and they unprotect. You depend on the parameters passed to them to determine what they do in any given call. If you keep the helpers, please add comments to each parameter explaining what effect it's going to have. But I would really recommend that you ditch the helper methods and have each top-level method do exactly what it's name says; e.g., "unprotectDataInWorksheet". You'll have a lot of repetitive code, but that's OK in snippets, even desirable.

samples/excel/89-preview-apis/protect-data-in-worksheet-and-workbook-structure.yaml
```diff
@@ -0,0 +1,258 @@
+id: excel-protect-data-in-worksheet-and-workbook-structure
+name: Protect data in worksheet and the workbook structure
+description: Shows how to protect data in a worksheet and the workbook structure.
+host: EXCEL
+api_set:
+    ExcelApi: 1.7
+script:
+    content: |
+        $("#setup").click(() => tryCatch(setup));
```
Rick-Kirkham: It looks like you've got a somewhat complex structure in order to avoid repetitive code. This is going to be problematic when we try to extract these functions for use in the reference topics. For example, your helper methods would have to be used for _both_ the `protect` and `unprotect` APIs, and the reader will have to figure out that the parameters passed to the method is what determines what it's going to do. 
Our Script Lab snippets should be relentlessly focused on showing clearly the Office APIs. All other considerations, including reusing code, are relatively unimportant. Devs get their best practice education elsewhere.

samples/excel/89-preview-apis/protect-data-in-worksheet-and-workbook-structure.yaml
```diff
@@ -18,100 +18,145 @@ script:
...
+                if (!activeSheet.protection.protected) {
+                    activeSheet.protection.protect();
+                }
+
+                activeSheet.load("name");
+                activeSheet.load("protection/protected");
+
+                await context.sync();
+
+                console.log("Expected: Worksheet '" + activeSheet.name + "' should " + messageHelper(true) + ". Protected? " + activeSheet.protection.protected);
```
Rick-Kirkham: The user can see visually that most of the buttons on the **Home** ribbon gray out when the sheet is protected and regain their color when the sheet is unprotected. So you don't really need the `console.log` at the end of the method. If you removed it, then you could also get rid of the `context.sync` just above it and the two `load` calls just above that.

samples/excel/89-preview-apis/protect-data-in-worksheet-and-workbook-structure.yaml
```diff
@@ -18,100 +18,145 @@ script:
...
-                await worksheetHelper(context, false, false);
+                let activeSheet = context.workbook.worksheets.getActiveWorksheet();
+                activeSheet.protection.unprotect();
+
+                activeSheet.load("name");
+                activeSheet.load("protection/protected");
+
+                await context.sync();
+
+                console.log("Expected: Worksheet '" + activeSheet.name + "' should " + messageHelper(false) + ". Protected? " + activeSheet.protection.protected);
```
Rick-Kirkham: Again, since the ribbon shows whether the sheet is unprotected, do you really need the console.log?

samples/excel/89-preview-apis/protect-data-in-worksheet-and-workbook-structure.yaml
```diff
@@ -18,100 +18,145 @@ script:
...
             });
         }
 
         async function protectWorkbookStructure() {
```
Rick-Kirkham: For the workbook structure functions, instead of having console.log that tells people what happened, give instructions in the page just above the button that tells them how to verify. E.g., have a brief paragraph that says "Click the next button and then notice that you cannot delete or add a new worksheet."

PR #144, opened 2018-02-21
Set the doughnutHoleSize property in a series for a dough nut chart.

siewmoi:


samples/excel/89-preview-apis/chart-series-doughnutholesize.yaml
```diff
@@ -0,0 +1,138 @@
...
+                let xRangeSelection = sheet.getRange("A1:A7");
+
+                // Add a series.
+                let newSeries = seriesCollection.series.add("Qtr2");
+                newSeries.setValues(rangeSelection);
+                newSeries.setXAxisValues(xRangeSelection);
+
+               // Set the size of the dough nut hole,
+               // in this case, zero, which means no hole.
+                newSeries.doughnutHoleSize = 0;
```
Rick-Kirkham: Seems odd to illustrate an API by setting the lowest possible value. Maybe you could have another method that sets it to a positive value so people can see the difference.

samples/excel/89-preview-apis/chart-series-doughnutholesize.yaml
```diff
@@ -0,0 +1,138 @@
...
+                let seriesCollection = sheet.charts.getItemAt(0);
+                let rangeSelection = sheet.getRange("C2:C7");
+                let xRangeSelection = sheet.getRange("A1:A7");
+
+                // Add a series.
+                let newSeries = seriesCollection.series.add("Qtr2");
+                newSeries.setValues(rangeSelection);
+                newSeries.setXAxisValues(xRangeSelection);
+
+               // Set the size of the dough nut hole,
```
Rick-Kirkham: Please add to the comment what the unit of measurement is. Are the holes measured in pixels?

PR #146, opened 2018-02-22
Get and set title substrings for a chart.

siewmoi:


samples/excel/89-preview-apis/chart-title-substring.yaml
```diff
@@ -0,0 +1,126 @@
...
+
+        async function changeTitleSubstring() {
+            await Excel.run(async (context) => {
+                let sheet = context.workbook.worksheets.getItem("Sample");
+                let chart = sheet.charts.getItemAt(0);
+             
+                // Get first seven characters of the title and highlight them to green.
+                let font = chart.title.getSubstring(0, 7).font;
+                font.color = "Green";       
+            });
```
Rick-Kirkham: Is there a reason for not ending the Excel.run with a context.sync the way we usually do?

PR #150, opened 2018-02-28
Shows how to format the vertical and horizontal axis in a chart.

siewmoi:


samples/excel/89-preview-apis/chart-axis-formatting.yaml
```diff
@@ -0,0 +1,209 @@
...
+                axis.majorTickMark = "Cross";
+                // Set the product name label location.
+                // In this case, the location is at 100 units.
+                axis.setCrossesAt(100);
+
+                await context.sync();
+            });
+        }
+
+        async function displayUnit() {
```
Rick-Kirkham: You have 2 displayUnit functions. Please delete one of them.

samples/excel/89-preview-apis/chart-axis-formatting.yaml
```diff
@@ -0,0 +1,189 @@
...
+
+        async function displayUnit(context: Excel.RequestContext) {
+            let sheet = context.workbook.worksheets.getItem("Sample");
+
+            let chart = sheet.charts.getItem("SalesChart");
+
+            let axis = chart.axes.valueAxis;
+            axis.displayUnit = "Thousands";
+
+            await context.sync();
```
Rick-Kirkham: Sorry I didn't see this before, but since the displayUnit is called from within a function that itself has a context.sync, I think the one inside displayUnit is unneeded. I think if you get rid of it (and remove the async keyword from the beginning of displayUnit, it should still work.

PR #154, opened 2018-04-03
Remove private-samples code snippets (and mapping file info)

kbrandl:
Remove content and mapping file info that I'd previously created for injecting code snippets into API Reference docs.


kbrandl:
@Rick-Kirkham -- can you please review this PR (since at least one approving review is required for PRs in this repo)?  I'm just removing some data that I previously added to test out our ability to auto-inject JS snippets into the API Reference docs in the same way that we do for TS snippets; Script Lab content/functionality isn't in any way affected by this PR.

PR #159, opened 2018-05-04
shows how to handle the Worksheet.onSelectionChanged event

Rick-Kirkham:


Rick-Kirkham:
This snippet was reviewed by Shiyuan in Feb but held from publication by a product bug. So I'm approving immediately.

PR #162, opened 2018-05-08
Implementing api name changes in excel snippets

Rick-Kirkham:


Rick-Kirkham:
I'm going to skip review because the changes are simple and this is urgent and potential reviewers are OOF or busy with //build conference.

PR #177, opened 2018-06-14
Mapping Word ScriptLab Snippets for Ref Gen

AlexJerabek:
word.xlsx maps the ScriptLab snippets to methods and properties. I encourage reviewers to look at that Excel file and ensure these are the most appropriate mappings. Thanks!


samples/word/02-paragraphs/insert-in-different-locations.yaml
```diff
@@ -27,7 +27,7 @@ script:
 
         async function start() {
             await Word.run(async (context) => {
-                // This button assumes the before was clicked, so we get the next paragraph and insert text at the begining.Note that there are no-valid locations depending on  the object. For instance insertParagraph and "before" on a paragrpah object is not a valid combination.
+                // This button assumes the before was clicked, so we get the next paragraph and insert text at the begining. Note that there are no-valid locations depending on the object. For instance insertParagraph and "before" on a paragraph object is not a valid combination.
```
ElizabethSamuel-MSFT: should be "beginning"

ElizabethSamuel-MSFT: instead of no-valid, perhaps "invalid"?

AlexJerabek: @ElizabethSamuel-MSFT: Fixed up the comments in those snippets. I also unmapped some of the ones in that files because they are redundant.

PR #178, opened 2018-06-18
update mappings

ElizabethSamuel-MSFT:


snippet-extractor-output/snippets.yaml
```diff
@@ -316,7 +316,7 @@ Excel.CustomXmlPartCollection.set:
         targetRange.format.autofitColumns();
         await context.sync();
     });
-Excel.CustomXmlPart.add:
+Excel.CustomXmlPartCollection.add:
```
AlexJerabek: Should this target CustomXMLPart.getXml instead?

snippet-extractor-output/snippets.yaml
```diff
@@ -482,7 +482,7 @@ Excel.CustomXmlPartScopedCollection.getItemOrNullObject:
 
         await context.sync();
     });
-Excel.CustomXmlPartScopedCollection.getByNamespace:
```
AlexJerabek: Maybe this should be a general example of Excel.CustomXmlPartScopedCollection (not an individual method).

snippet-extractor-output/snippets.yaml
```diff
@@ -512,7 +512,7 @@ Excel.CustomXmlPartScopedCollection.getByNamespace:
 
         await context.sync();
     });
-Excel.Worksheet.getCount:
+Excel.CustomXmlPartCollection.getCount:
```
AlexJerabek: Should this be CustomXmlPartScopedCollection?

snippet-extractor-output/snippets.yaml
ElizabethSamuel-MSFT: The same snippet is already/also tagged for CustomXMLPart.getXml

snippet-extractor-output/snippets.yaml
ElizabethSamuel-MSFT: the property getByNamespace is of type Excel.CustomXmlPartScopedCollection but belongs to Excel.CustomXmlPartCollection which is why I've set it like that. If anything, there could be an argument for setting the example at Excel.CustomXmlPartCollection level.

AlexJerabek: Right. I guess my thought was this snippet is "some stuff done to a scoped collection", so it should be there. I don't have a strong opinion, so I trust your judgement.

ElizabethSamuel-MSFT: Actually, I should say it returns a Scoped one.

PR #183, opened 2018-08-06
Chart formatting example

AlexJerabek:
An example showing how to use formatting properties with a chart. This is intended to over the 1.8 API chart scenarios.

### DIFF
--- a/samples/excel/89-preview-apis/chart-formatting.yaml
+++ b/samples/excel/89-preview-apis/chart-formatting.yaml
@@ -29,7 +29,7 @@ script:
                 chart.title.text = "GDP Per Capita , 1980 to 2017";
                 chart.legend.position = "Bottom";
 
-                worksheet.onChanged.add(onDataChanged);
+                await context.sync();
             });
         }
 
@@ -54,6 +54,7 @@ script:
                 chart.axes.categoryAxis.format.font.color = "#D9D9D9";
                 chart.axes.valueAxis.majorGridlines.format.line.color = "#585858";
                 chart.format.fill.setSolidColor("#404040");
+                await context.sync();
             });
         }
 
@@ -91,6 +92,7 @@ script:
                 chart.dataLabels.position = "Center";
                 chart.dataLabels.separator = "\n";
                 chart.dataLabels.format.font.color = "#D9D9D9";
+                await context.sync();
             });
         }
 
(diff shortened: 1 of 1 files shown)


samples/excel/89-preview-apis/chart-formatting.yaml
```diff
@@ -0,0 +1,328 @@
...
+                chart.top = 275;
+                chart.left = 50;
+                chart.height = 600;
+                chart.width = 850;
+                
+                chart.title.text = "GDP Per Capita , 1980 to 2017";
+                chart.legend.position = "Bottom";
+
+                worksheet.onChanged.add(onDataChanged);
+            });
```
Rick-Kirkham: Its a good practice to always have one last `await context.sync()` at the end of an `Excel.run`, even though the infrastructure will do this for you if you don't. It makes it clear to someone reading the code that there is a sync at that point. There's no danger of a redundant sync because the infrastructure will not call its own sync if the last line of the `run `calls it explicitly.

samples/excel/89-preview-apis/chart-formatting.yaml
```diff
@@ -0,0 +1,328 @@
...
+                chart.title.format.font.color = "#D9D9D9";
+
+                chart.legend.format.font.name = "Times New Roman";
+                chart.legend.format.font.color = "#D9D9D9";
+
+                chart.axes.categoryAxis.format.font.name = "Times New Roman";
+                chart.axes.categoryAxis.format.font.color = "#D9D9D9";
+                chart.axes.valueAxis.majorGridlines.format.line.color = "#585858";
+                chart.format.fill.setSolidColor("#404040");
+            });
```
Rick-Kirkham: The point about having one last `await context.sync()` applies here, too; and to all the Excel.runs in the file.

samples/excel/89-preview-apis/chart-formatting.yaml
```diff
@@ -0,0 +1,328 @@
...
+                await context.sync();
+                
+                for (var i = 0; i < charts.count; i++) {
+                    charts.getItemAt(0).delete();
+                }
+            });
+        }
+
+        async function addSampleData() {
+            try {
```
Rick-Kirkham: I don't think the try/catch _inside_ this method is needed because the whole `addSampleData `is passed to the `tryCatch()` method.

AlexJerabek:
Thanks @Rick-Kirkham! I made the requested changes. I also removed the onDatChanged listener, as it wasn't being used in this samples. I'll make a separate snippet for 1.8 chart events.

PR #186, opened 2018-08-15
Event toggling snippet

AlexJerabek:
This sample shows how to turn on and off events.

My intention is to make this part of a larger sample once calculation toggling goes online. Hence why there's more code and data here than absolutely necessary.

### DIFF
--- a/samples/excel/89-preview-apis/events-disable-events.yaml
+++ b/samples/excel/89-preview-apis/events-disable-events.yaml
@@ -1,0 +1,166 @@
+id: excel-events-disable-events
+name: Disable Events
+description: Toggle event firing on and off for performance.
+host: EXCEL
+api_set:
+    ExcelAPI: 1.8
+script:
+    content: |+
+        $("#toggleEvents").click(() => tryCatch(toggleEvents));
+        $("#setup").click(() => tryCatch(setup));
+        $("#refreshData").click(() => tryCatch(addOrRefreshData));
+        $("#registerSumChangedHandlers").click(() => tryCatch(registerSumChangedHandlers));
+
+        async function toggleEvents() {
+            await Excel.run(async (context) => {
+                context.runtime.load("enableEvents");
+                await context.sync();
+                const eventBoolean = !context.runtime.enableEvents
+                context.runtime.enableEvents = eventBoolean;
+                if (eventBoolean) {
+                    console.log("Events are currently on.");
+                } else {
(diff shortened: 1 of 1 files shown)


samples/excel/89-preview-apis/events-disable-events.yaml
```diff
@@ -0,0 +1,166 @@
...
+            catch (error) {
+                OfficeHelpers.UI.notify(error);
+                OfficeHelpers.Utilities.log(error);
+            }
+        }
+    language: typescript
+template:
+    content: |-
+        <section class="ms-font-m">
+            <p>This sample shows how to turn on and off events.</p>
```
ElizabethSamuel-MSFT: maybe "turn events on and off" instead

samples/excel/89-preview-apis/events-disable-events.yaml
```diff
@@ -0,0 +1,166 @@
...
+                // Log the address of the sum that changed
+                const range = context.workbook.bindings.getItem(eventArgs.binding.id).getRange();
+                const currentDate = new Date();
+                const cell = range.getCell(1, 0);
+                cell.formulas = [["=TIME(" + currentDate.getHours() + "," + currentDate.getMinutes() + "," + currentDate.getSeconds() + ")"]];
+                await context.sync();
+            });
+        }
+
+        async function setup() {
```
ElizabethSamuel-MSFT: I recommend that you have setup create a new sheet with some appropriate data. Then you don't have to worry about interfering with the user's sheets and the other methods can look for that sheet.

AlexJerabek:
@ElizabethSamuel-MSFT Thanks for the feedback. Updated accordingly.

I also changed what the event handler does to be more obvious when it's on or off.

samples/excel/89-preview-apis/events-disable-events.yaml
```diff
@@ -0,0 +1,177 @@
...
+            });
+        }
+
+        async function onSumChanged(eventArgs: Excel.BindingDataChangedEventArgs) {
+            await Excel.run(async (context) => {
+                const sheet = context.workbook.worksheets.getItem("Sample");
+                const cell = sheet.getRange("P21");
+                 
+                // get the grand total of the whole sum range
+                // note that we are using having this program perform the sum instead of giving the Excel cell a function,
```
ElizabethSamuel-MSFT: Should this start with: // note that we are having

PR #187, opened 2018-08-17
Copy and paste sample

AlexJerabek:
A sample showcasing Range.copyFrom(), aka copy and paste.

### DIFF
--- a/samples/excel/89-preview-apis/range-copyfrom.yaml
+++ b/samples/excel/89-preview-apis/range-copyfrom.yaml
@@ -1,6 +1,6 @@
 id: excel-range=copyfrom
 name: Copy and paste ranges
-description: Use the range.copyFrom method to copy data and formatting.
+description: Use the Range.copyFrom method to copy data and formatting.
 host: EXCEL
 api_set:
     ExcelAPI: 1.8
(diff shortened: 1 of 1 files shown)


samples/excel/89-preview-apis/range-copyfrom.yaml
```diff
@@ -0,0 +1,194 @@
...
+            await Excel.run(async (context) => {
+                const sheet = context.workbook.worksheets.getItem("Sample");
+                sheet.getRange("F6").values = [["Copied Without Blanks"]];
+                sheet.getRange("F6").format.autofitColumns();
+
+                // fill the destination range so we can see the blank being skipped
+                sheet.getRange("G6:K6").values = [["Old Data", "Old Data", "Old Data", "Old Data", "Old Data"]]
+
+                // copy a range, omitting the blank cells so existing data is not overwritten in those cells
+                sheet.getRange("G6").copyFrom("A1:E1", Excel.RangeCopyType.all, true, false);
```
Rick-Kirkham: Since Boolean parameters aren't self-documenting its a good idea to put the parameter name in an inline comment. So this line would read:
`eet.getRange("G6").copyFrom("A1:E1", Excel.RangeCopyType.all, true /* skipBlanks */, false /* transpose */); `

PR #190, opened 2018-08-23
Script lab linking

AlexJerabek:


playlists/excel.yaml
```diff
@@ -526,8 +526,8 @@
   fileName: refresh-pivot-table.yaml
   description: Refresh pivot table
   rawUrl: >-
-    https://raw.githubusercontent.com/<ACCOUNT>/<REPO>/<BRANCH>/samples/excel/55-pivot-table/refresh-pivot-table.yaml
-  group: Pivot Table
+    https://raw.githubusercontent.com/<ACCOUNT>/<REPO>/<BRANCH>/samples/excel/55-pivottable/refresh-pivot-table.yaml
+  group: Pivottable
```
kbrandl: Interesting that this got changed to "Pivottable" (lowercase "t") by the build script, as it seems like (IIRC) the value of the `Group` attribute is what is displayed in the samples menu. Suggest that you verify the Samples menu (in Script Lab beta) once you merge this PR to master.  If the value there is "Pivottable", then the fact that the build script changed this attribute value like this (i.e., to no longer be camel-cased) seems like a bug in the build script that we should fix.

AlexJerabek: @kbrandl Yeah, is displays as "Pivottable" in the samples menu. I think it's because the folder name isn't camel-cased, but the script doesn't allow non-lowercased folder names. Should I change that?

PR #204, opened 2018-09-10
Correcting sample to follow product team guidance

AlexJerabek:
Here's the email text from Jipeng:

@Alexander Jerabek For the createworkbook API, the context.application.createWorkbook().open(); is not suggested to be used by user.

Instead we have changed to the following way, where can we put the documentation? Thanks.

Excel.createWorkbook(base64).catch(function (error) {
                app.showNotification(error.code, error.message);
                if (error instanceof OfficeExtension.Error) {
                    app.showNotification("Debug info: ", JSON.stringify(error.debugInfo));
                }
            });

### DIFF
--- a/samples/excel/82-document/get-file-in-slices-async.yaml
+++ b/samples/excel/82-document/get-file-in-slices-async.yaml
@@ -83,11 +83,11 @@ script:
             }
         }
 
-        // Note: this function uses preview APIs. Switch to the beta libraries to use Application.createWorkbook().
+        // Note: this function uses preview APIs. Switch to the beta libraries to use CreateWorkbook().
         async function newWorkbookFromFile() {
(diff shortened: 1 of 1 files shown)


AlexJerabek:
@Rick-Kirkham The Travis-ci check has been pending for hours. Is there a way to kick that off again? Alternatively, is there a way to set a stricter timeout? I believe Travis times out on DefinitelyTyped in ~15 minutes.

Rick-Kirkham:
@AlexJerabek All I know about it is this line from the readme for this repo: "Once it a while, the online test fails when the local test passed. This is usually a bug in the online test service. If this happens, cancel the pull request, wait a few hours, and then repeat the steps for creating a pull request."

PR #213, opened 2018-10-09
Adding event snippet to mapping and running npm start

AlexJerabek:
FYI, a lot of this code churn is due to alphabetizing the excel file.


AlexJerabek:
Here's the excel file addition:


Class | Method/Prop/Rel Name | SnippetIdIntheYAMLFile | MethodNameInTheSnippet
-- | -- | -- | --
Runtime | enableEvents | excel-events-disable-events | toggleEvents

PR #216, opened 2018-10-24
Fix for jQuery Intellisense in Script Lab

nico-bellante:
This is a short-term fix for [script-lab-react](https://github.com/OfficeDev/script-lab-react), and a fix for [script-lab](https://github.com/OfficeDev/script-lab).

This issue, documented [here](https://github.com/OfficeDev/script-lab-react/issues/170), stems from jQuery changing its types to be split out into multiple files instead of just one `index.d.ts` file. Script Lab currently assumes all types will be in the `index.d.ts` file and does not properly handle triple slash references.

image

This change sets all samples to explicitly request `@types/jQuery@3.1.1` instead of leaving the version blank and defaulting to the latest version in which these breaking changes for Script Lab occur.

### DIFF
--- a/private-samples/excel/20-chart/chart-title-js.yaml
+++ b/private-samples/excel/20-chart/chart-title-js.yaml
@@ -149,4 +149,4 @@ libraries: |
     @types/office-js
     @types/core-js
     @microsoft/office-js-helpers@0.7.4/dist/office.helpers.d.ts
-    @types/jquery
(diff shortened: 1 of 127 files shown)


Zlatkovsky:
@AlexJerabek , could you create a PR for Master->Prod, assuming that master is in a good shape to move to Prod?  Or, if not, cherry-pick this commit onto the prod branch?

PR #229, opened 2018-11-14
Excel Snippet Work

AlexJerabek:
- Standardize try/catch pattern
- Automate flow of Test for used ranges
- Explain failure conditions better in Table->Style
- Add rows and columns now adds to existing table rather than resetting it
- Consolidate named item snippets
- Clean up Chart Series sample
- Rework Chart Title Substring into more general Chart Title snippet
- Remove obsolete PivotTable sample
- Remove failure states from custom xml sample
- Merge event snippets


samples/excel/30-events/events-worksheet.yaml
```diff
@@ -0,0 +1,202 @@
+order: 8
+id: excel-events-worksheet
+name: Worksheet events
+description: 'Registers event handlers for selection, data, and calculation changes.'
```
kbrandl: For consistency with other snippet descriptions, change to -> `Registers event handlers that run when data is changed in worksheet, the selected range changes in a worksheet, or the worksheet is recalculated.`

samples/excel/30-events/events-workbook-and-worksheet-collection.yaml
```diff
@@ -0,0 +1,257 @@
+order: 7
+id: excel-events-workbook-and-worksheet-collection
+name: Workbook and WorksheetCollection events
+description: 'Registers event handlers for worksheet addition, activation, deactivation, and settings changes.'
```
kbrandl: For consistency with other snippet descriptions, change to -> `Registers event handlers that run when a worksheet is added, activated, or deactivated, or worksheet settings are changed.`

PR #231, opened 2018-11-15
Snippet standization cleanup

AlexJerabek:
Reverting these snippets back to the aforementioned pattern. It should be noted ScriptLab is what stripped them.


AlexJerabek:
Closing to respond to more feedback given offline.

AlexJerabek:
Responded to @kbrandl's offline feedback. Cleaned up descriptions, whitespace, and UI.notify (those go off when there's an error-like condition, whereas console.log displays information).

Also, the Table events works fine for me. What enivronment are you using?

PR #232, opened 2018-11-15
CreateWorkbook snippet

AlexJerabek:
Based on @JuaneloJuanelo's [example snippet](https://gist.github.com/JuaneloJuanelo/399608d6928f6183b474af8f1f7e1c8a) provided for a partner. 

The sample has a file chooser which, upon selection, reads the file as a base64-encoded string and creates a new workbook based on that string.


samples/excel/50-workbook/create-workbook.yaml
```diff
@@ -0,0 +1,86 @@
...
+    ExcelApi: 1.8
+script:
+    content: |
+        $("#file").change(() => tryCatch(createWorkbook));
+
+        async function createWorkbook() {
+            const myFile = document.getElementById("file");
+            const reader = new FileReader();
+
+            reader.onload = ((theFile) => {
```
Rick-Kirkham: Why are we assigning an IFFE to reader.onload? Examples I find online for reader.onload don't do this. Maybe I'm missing something, but this looks like its going to pointlessly call the function immediately after it's assigned to onload, even before the readAsDataURL line runs, and then run again when the latter line runs.

samples/excel/50-workbook/create-workbook.yaml
```diff
@@ -0,0 +1,86 @@
...
+                    Excel.run(context => {
+                        // strip off the metadata before the base64-encoded string
+                        const startIndex = event.target.result.indexOf("base64,");
+                        const mybase64 = event.target.result.substr(startIndex + 7);
+
+                        Excel.createWorkbook(mybase64);
+                        return context.sync();
+                    });
+                });
+            })(myFile.files[0]);
```
Rick-Kirkham: I get an error from myFile.files[0] saying that HTMLElement does not have a property named files.

samples/excel/50-workbook/create-workbook.yaml
```diff
@@ -0,0 +1,86 @@
...
+                await callback();
+            } catch (error) {
+                OfficeHelpers.UI.notify(error);
+                OfficeHelpers.Utilities.log(error);
+            }
+        }
+    language: typescript
+template:
+    content: |-
+        <section class="ms-font-m">
```
Rick-Kirkham: I don't understand what this sample is supposed to do. the only UI is a browse control. I browse to an Excel file and select it. But then nothing happens. Is something supposed to happen?

samples/excel/50-workbook/create-workbook.yaml
AlexJerabek: I do too, but that's how that API works ([SO example here](https://stackoverflow.com/questions/8645369/how-do-i-get-the-file-content-from-a-form)). Am I supposed to import something?

samples/excel/50-workbook/create-workbook.yaml
AlexJerabek: It's supposed to open a copy of the workbook in a new instance. What environment are you running this snippet in?

AlexJerabek:
Thanks for taking a look @Rick-Kirkham and @kbrandl. I've updated the code and replied to particular items.

samples/excel/50-workbook/create-workbook.yaml
Rick-Kirkham: @AlexJerabek Office 365 1812 11113.20002 Click to Run. I've tried both prod and beta Script Lab.

samples/excel/50-workbook/create-workbook.yaml
Rick-Kirkham: @AlexJerabek  You have to do some casting. The same problem and solution applies to EventTarget.result. Here's what the code should look like:

```
const myFile = <HTMLInputElement>document.getElementById("file");
    const reader = new FileReader();
    reader.onload = ((event) => {
        Excel.run(context => {
            // strip off the metadata before the base64-encoded string 
            const startIndex = (<FileReader>event.target).result.indexOf("base64,");
            const mybase64 = (<FileReader>event.target).result.substr(startIndex + 7);
            Excel.createWorkbook(mybase64);
            return context.sync();
        });
```

samples/excel/50-workbook/create-workbook.yaml
```diff
@@ -0,0 +1,94 @@
...
+            <p>This sample shows how to create a new workbook by copying an existing one.</p>
```
ElizabethSamuel-MSFT: Seems like, based on additional scenario, this paragraph should be updated

playlists/excel.yaml
```diff
@@ -513,6 +513,15 @@
   group: Workbook
   api_set:
     ExcelApi: 1.7
+- id: excel-workbook-create-workbook
+  name: Create workbook
+  fileName: create-workbook.yaml
+  description: Sample for Excel
```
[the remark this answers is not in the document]
kbrandl: @AlexJerabek FYI, the `description` still needs to be updated.

samples/excel/50-workbook/create-workbook.yaml
```diff
@@ -0,0 +1,94 @@
+order: 2
+id: excel-workbook-create-workbook
+name: Create workbook
+description: Creates new workbooks both from scratch and by copying an existing one.
```
kbrandl: Consider changing description to something more like -> "Creates a new (empty) workbook and creates a new workbook by copying an existing one."

samples/excel/50-workbook/create-workbook.yaml
```diff
@@ -0,0 +1,94 @@
...
+    content: |-
+        <section class="ms-font-m">
+            <p>This sample shows how to create a new workbook by copying an existing one.</p>
+        </section>
+
+        <section class="samples ms-font-m">
+            <h3>Try it out</h3>
+            <button id="create-new-blank-workbook" class="ms-Button">
+                <span class="ms-Button-label">Create new blank workbook</span>
+            </button><p>
```
kbrandl: Suggest that you add a heading and a sentence above the button (to make it clearer visually the the snippet lets you do two different things -- 1) create empty workbook, 2) copy existing workbook) and change the button label to `Create workbook`

```
<p><b>Create empty workbook</b></p>
<p>Press the button to create a new (empty) workbook and open it in a new instance of Excel.</p>
<button id="create-new-blank-workbook" class="ms-Button">
    <span class="ms-Button-label">Create workbook</span>
</button><p>

samples/excel/50-workbook/create-workbook.yaml
kbrandl: +1 Perhaps something like -> "This sample shows how to create a new (empty) workbook and how to create a new workbook by copying an existing one."

samples/excel/50-workbook/create-workbook.yaml
AlexJerabek: Added a new header. I like the pattern of a header per distinct scenario. I found adding more text to be redundant. The button is self-explanatory, especially with the snippet description.

AlexJerabek:
Thank you @ElizabethSamuel-MSFT, @kbrandl, and @Rick-Kirkham. Updated according to feedback.

PR #239, opened 2018-12-05
[Word] (Range) Point snippet at correct method

AlexJerabek:
The sample was being pointed at Paragraph.getTextRanges, instead of Range.getTextRanges.


snippet-extractor-output/snippets.yaml
```diff
@@ -2533,7 +2533,7 @@ Word.Paragraph.text:
         await context.sync();
         console.log(paragraph.text);
     });
-Word.Paragraph.getTextRanges:
+Word.Range.getTextRanges:
   - |-
     await Word.run(async (context) => {
         // Gets the complete sentence  (as range) associated with the insertion point.
```
ElizabethSamuel-MSFT: nit: could you clean up the spacing here?

AlexJerabek:
Thanks @ElizabethSamuel-MSFT. I cleaned up the spacing. Was that what you had in mind?

PR #242, opened 2018-12-21
[excel] (Chart) Clean up chart formatting snippet

AlexJerabek:
The output of this snippet always looked jarringly different than the others. Cleaned it up to use a similar set of sample data to the other chart snippets.

Also fixed a small bug in the chart axis formatting snippet and re-ran NPM start.


kbrandl:
@AlexJerabek looks like this PR also contains a new snippet (**excel-range-remove-duplicates**) -- is that expected?

AlexJerabek:
@kbrandl The remove duplicates snippet was checked in with PR #240.  It's only showing up here because of some syncing problem.

As for the console numbers being off, that's due to the fact blank rows are counted (as [documented here](https://github.com/OfficeDev/office-js-snippets/pull/240)). I'll adjust the snippet text accordingly.

PR #244, opened 2018-12-28
[excel] (Range) Adding find snippet

AlexJerabek:
This snippet covers the usage of `Range.find`. Users can search a table for text they enter in a text area and toggle various search criteria.


samples/excel/85-preview-apis/range-find.yaml
```diff
@@ -0,0 +1,171 @@
...
+                const sheet = context.workbook.worksheets.getItem("Sample");
+                const table = sheet.tables.getItem("ExpensesTable");
+                const searchRange = table.getRange();
+                const foundRange = searchRange.find($("#searchText").text(), {
+                    completeMatch: completeMatch,
+                    matchCase: matchCase,
+                    searchDirection: searchDirection
+                });
+                
+                foundRange.load("address");
```
kbrandl: Consider adding a code comment before this line to indicate that an error will be thrown if no match is found. This'll help the user to understand that the ItemNotFound error is by design (not a bug with the snippet). Maybe something along the lines of: 

`// Note: If no match was found, an ItemNotFound error will be thrown when this line attempts to load the address property.`

AlexJerabek:
Thanks @kbrandl! I added the note. It's actually the `Range.find` line throwing `ItemNotFound` when there's no match (I did some testing), so I put the note there.

PR #245, opened 2018-12-28
[excel] (Worksheet) Find all

AlexJerabek:
Adding a Worksheet.findAll snippet. This shows how to search for strings across a worksheet and quickly edit the returned `RangeAreas`.


samples/excel/85-preview-apis/worksheet-find-all.yaml
```diff
@@ -0,0 +1,154 @@
...
+            } catch (error) {
+                OfficeHelpers.UI.notify(error);
+                OfficeHelpers.Utilities.log(error);
+            }
+        }
+    language: typescript
+template:
+    content: |-
+        <section class="ms-font-m">
+            <p>This sample shows how to find strings across an entire worksheet.</p>
```
kbrandl: For consistency with snippet title/description (and with the other 'find' snippet you created previously) -- change "how to find strings..." -> "how to find data..."

samples/excel/85-preview-apis/worksheet-find-all.yaml
```diff
@@ -0,0 +1,154 @@
+order: 4
+id: excel-worksheet-find-all
+name: Find all ranges
```
kbrandl: Even though this name describes what the API is doing (it's finding ranges), it doesn't sound like what the snippet does, from a functional perspective. Snippet names should describe what the snippet does from a functional perspective, because that's how the user is likely to think of things (i.e., a user is more likely to think in terms of "I want to find data in a worksheet" than "I want to find all ranges"). 

And, I just noticed that the previous snippet (`excel-range-find`) is named **Find** -- probably should rename that snippet as well.

- Consider renaming this snippet to "Find data in worksheet"
- Consider renaming the `excel-range-find` snippet to "Find data in range"

samples/excel/85-preview-apis/worksheet-find-all.yaml
```diff
@@ -0,0 +1,154 @@
...
+                    foundRanges.format.fill.color = "yellow"
+                }
+            });
+        }
+
+        async function findCanceled() {
+            await Excel.run(async (context) => {
+                const sheet = context.workbook.worksheets.getItem("Sample");
+
+                // NOTE: In this sample, there are no canceled projects. Using findAll (without the OrNullObject) would throw an ItemNotFound error.
```
kbrandl: Consider rewording -> 

// NOTE: In this sample, there are no canceled projects. Calling sheet.findAll(...) instead of sheet.findAllOrNullObject(...) would throw an ItemNotFound error.

PR #249, opened 2019-01-08
Updating default snippets

AlexJerabek:
This changes the default snippets in the following ways:
- `console.log` is used instead of `OfficeHelpers.UI.notify`.
- The style is the same as all the other snippets.
- Excel and Word get a basic call to get the object they're most likely to use.


samples/excel/default.yaml
```diff
@@ -11,8 +11,9 @@ script:
 
         async function run() {
             await Excel.run(async (context) => {
+                const sheet = context.workbook.worksheets.getActiveWorksheet();
 
-                OfficeHelpers.UI.notify("Your code goes here");
+                console.log("Your code goes here...");
```
Zlatkovsky: Why the "..." here but not in the others?
Oh, I see, that was what I actually had copy-pasted for you.  My bad.  I didn't want the "..."s there.  I suggest we remove them.

samples/outlook/default.yaml
```diff
@@ -9,7 +9,7 @@ script:
         $("#run").click(run);
 
         function run() {
-            OfficeHelpers.UI.notify(`Hello ${Office.context.mailbox.userProfile.displayName}`);
+            console.log(`Hello ${Office.context.mailbox.userProfile.displayName}`);
```
Zlatkovsky: To be consistent with the other samples (and make deleting stuff easier), should we make this a two-liner?

```
const displayName = Office.context.mailbox.userProfile.displayName;
console.log(`Hello ${displayName}`);
```

AlexJerabek:
Thanks for the feedback @Zlatkovsky. I've updated those lines accordingly.

PR #256, opened 2019-01-28
Increasing accuracy of ChartCollection event snippet description

AlexJerabek:
@jakobpn, aside from this description needing changes, were there other issues you noticed with the chart event snippets?

### DIFF
--- a/playlists/excel.yaml
+++ b/playlists/excel.yaml
@@ -217,8 +217,9 @@
   name: Chart collection events
   fileName: events-chartcollection-added-activated.yaml
   description: >-
-    Registers event handlers that run when a chart collection is activated,
-    deactivated, added, or deleted.
+    Registers event handlers on a worksheet's chart collection that run when any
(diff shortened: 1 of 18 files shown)


jakobpn:
Hi Alex, thx for looking into this. To answer your question, yes, there are currently 2 snippets for chart events. To me it looks like they have overlapping functionality and I suggest that is rationalized so there is either just one snippet or 2 snippets where it is clear which one to use when.

AlexJerabek:
Hi @jakobpn,

The snippets are targeting the events of the [Chart](https://docs.microsoft.com/en-us/javascript/api/excel/excel.chart?view=office-js#events) and [ChartCollection](https://docs.microsoft.com/en-us/javascript/api/excel/excel.chartcollection?view=office-js) objects. They're meant to demonstrate behavioral differences between targeting the individual chart and the whole collection. 

I've updated the descriptions to better differentiate between the collection and individual. Thoughts?

PR #262, opened 2019-02-15
[all] (ScriptLab UI) Switching out OfficeHelpers.UI.notify/log in favor of console.log/error

AlexJerabek:
Due to the improvements made to Script Lab's console, error objects have specialized display support. We'll favor using that console going forward.


samples/excel/30-events/events-workbook-and-worksheet-collection.yaml
```diff
@@ -136,7 +136,7 @@ script:
                 });
             }
             catch (error) {
-                OfficeHelpers.UI.notify(error);
+                console.log(error);
```
ElizabethSamuel-MSFT:
```suggestion
                console.error(error);
```

PR #274, opened 2019-03-13
[excel] (Comments) Adding snippets for comment APIs

AlexJerabek:


kbrandl:
@AlexJerabek sorry for the slow response on this one -- would it be possible for you to have someone else review it? I'm OOF for the remainder of the week, starting in another half hour, and rushing to wrap up other things before then. I'd be happy to review it on Monday when I'm back in the office, if you're still awaiting review then. Thanks!

samples/excel/85-preview-apis/comment.yaml
```diff
@@ -0,0 +1,167 @@
...
+            await Excel.run(async (context) => {
+                const sheet = context.workbook.worksheets.getItem("Comments");
+                sheet.comments.add("TODO: add data.", "A2");
+                await context.sync();
+            });
+        }
+        async function addCommentReply() {
+            await Excel.run(async (context) => {
+                const comment = context.workbook.comments.getItemByCell("Comments!A1");
+                context.load(comment);
```
Rick-Kirkham: I'd like to see a code comment explaining why the comment object has to be loaded before you can add to its replies collection. Also, `context.load(comment)` is equivalent to `comment.load()', so its an empty load that causes all of the comment's scalar properties to be loaded. Is that really necessary?

samples/excel/85-preview-apis/comment.yaml
```diff
@@ -0,0 +1,167 @@
+order: 1
+id: excel-comment
+name: Comments
+description: 'Adds, edits, and removes comments and comment replies.'
+host: EXCEL
+api_set:
+    ExcelApi: 1.9
+script:
+    content: |
+        $("#setup").click(() => tryCatch(setup)); $("#addCommentToSelectedCell").click(() => tryCatch(addCommentToSelectedCell)); $("#addCommentToCell").click(() => tryCatch(addCommentToCell)); $("#addCommentReply").click(() => tryCatch(addCommentReply)); $("#getCommentMetadata").click(() => tryCatch(getCommentMetadata)); $("#editComment").click(() => tryCatch(editComment)); $("#editCommentReply").click(() => tryCatch(editCommentReply)); $("#deleteComment").click(() => tryCatch(deleteComment)); $("#deleteCommentReply").click(() => tryCatch(deleteCommentReply));
```
Rick-Kirkham: Nit: Doesn't seem to be a line break at the semi-colons in the button handler assignments.

AlexJerabek:
Thanks @Rick-Kirkham. The first load statement in addCommentReply isn't actually necessary. Good catch. I've removed it and updated the PR accordingly.

PR #275, opened 2019-03-19
[admin] (Preview APIs) Switching beta types from the CDN to DT

AlexJerabek:


config/build.ts
```diff
@@ -234,9 +234,10 @@ async function processSnippets(processedSnippets: Dictionary<SnippetProcessedDat
         const officeDtsReferences =
             snippet.libraries.split('\n')
                 .map(reference => reference.trim())
-                .filter(reference => reference.match(/.*((@types\/office-js)|(office\.d\.ts))$/gi));
+                .filter(reference => reference.match(/.*((@types\/office-js(-preview)?)|(office\.d\.ts))$/gi));
```
Zlatkovsky: Do you want to add logic to enforce that the right d.ts is specified for the corresponding office.js reference -- just like we enforce office.js references to be of the canonical type?

Zlatkovsky: What I'm thinking is just somewhere after here, for you to compare `officeJsReferences` and `officeDtsReferences` and ensure that if one is official or beta, than so is the other.

AlexJerabek: I think we already do that on lines 269-271.

Zlatkovsky: Oh, sweet.  I didn't look far enough down.  Sounds like you're all good, then.

PR #280, opened 2019-03-26
[Word] Add Get word count snippet

ElizabethSamuel-MSFT:


samples/word/25-paragraph/get-word-count.yaml
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        async function setup() {
+          await Word.run(async (context) => {
+            // Let's insert a couple of paragraphs to illustrate the point.
+            context.document.body.clear();
+            context.document.body.insertParagraph("Video provides a powerful way to help you prove your point. When you click Online Video, you can paste in the embed code for the video you want to add. You can also type a keyword to search online for the video that best fits your document.", "Start");
+            context.document.body.insertParagraph("To make your document look professionally produced, Word provides header, footer, cover page, and text box designs that complement each other. For example, you can add a matching cover page, header, and sidebar. Click Insert and then choose the elements you want from the different galleries.", "End");
+
+            await context.sync();
```
Rick-Kirkham: This setup method seems to work even if this context.sync is commented out. Is it needed for some reason that I don't see?

samples/word/25-paragraph/get-word-count.yaml
```diff
@@ -0,0 +1,131 @@
...
+
+            context.document.body.paragraphs.getFirst().alignment = "Left";
```
Rick-Kirkham: What are these two lines for? I think paragraphs in Word are left-aligned by default.

samples/word/25-paragraph/get-word-count.yaml
```diff
@@ -0,0 +1,131 @@
...
+                paragraph.split(" ").forEach((term) => {
+                  let currentTerm = term.trim();
+                  if (currentTerm) {
+                    text.push(currentTerm);
+                  }
+                });
+              }
+            });
+
+            let makeTextDistinct = new Set(text);
```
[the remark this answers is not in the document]
ElizabethSamuel-MSFT: @Rick-Kirkham: Reference - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

samples/word/25-paragraph/get-word-count.yaml
ElizabethSamuel-MSFT: I grabbed the setup content from another Word snippet. Depending on what we decided here, I'll clean those up too.

PR #288, opened 2019-04-03
simplified custom function code

davidchesnut:


samples/excel/16-custom-functions/volatile-function.yaml
```diff
@@ -7,18 +7,14 @@ api_set:
...
-            roll = Math.round(roll);
-            result += roll;
-          }
-          return result;
+        /** @CustomFunction 
+        *  @description Simulates rolling a 6 sided die.
+        * @returns A whole number from 1 to 6.
+        * @volatile
+        */
+        function roll6sided(): number {
```
AlexJerabek: The cool kids would call this `rollD6`

samples/excel/16-custom-functions/volatile-function.yaml
```diff
@@ -7,18 +7,14 @@ api_set:
...
+          const roll: number = Math.random() * 5 + 1;
```
AlexJerabek: Nit: we usually don't explicitly define the type in our samples.

Zlatkovsky: +1.  Typescript will figure it out, so you don't need it.

In fact, you could just simplify it to `return Math.round(Math.random() * 5 + 1)` as a one-liner, and I think it would still be perfectly readable

samples/excel/16-custom-functions/volatile-function.yaml
```diff
@@ -7,18 +7,14 @@ api_set:
...
-            var roll:number;
-            roll = Math.random()*6;
-            roll = Math.round(roll);
-            result += roll;
-          }
-          return result;
+        /** @CustomFunction 
+        *  @description Simulates rolling a 6 sided die.
+        * @returns A whole number from 1 to 6.
+        * @volatile
```
Zlatkovsky: So does `@volatile` work now?

samples/excel/16-custom-functions/volatile-function.yaml
Zlatkovsky: I guess I'm not cool, so I wouldn't have known.  I think "roll6sided" is totally fine, to be inclusive of the uncool kids :-)

samples/excel/16-custom-functions/volatile-function.yaml
```diff
@@ -7,18 +7,14 @@ api_set:
...
-          result = 0;
-          for (i = 0;i<x;i++) {
-            var roll:number;
-            roll = Math.random()*6;
-            roll = Math.round(roll);
-            result += roll;
-          }
-          return result;
+        /** @CustomFunction 
+        *  @description Simulates rolling a 6 sided die.
```
ElizabethSamuel-MSFT: nit: 6-sided

samples/excel/16-custom-functions/volatile-function.yaml
davidchesnut: @Zlatkovsky I'm not sure. I use manual recalc to check that the functions are called again (which I understand is the volatile behavior) but they always work even if I leave volatile off. However, I notice script lab adds the volatile keyword to the metadata if I use it so I put it in.

samples/excel/16-custom-functions/volatile-function.yaml
davidchesnut: I like the one liner. Changed!

samples/excel/16-custom-functions/volatile-function.yaml
davidchesnut:
```suggestion
        *  @description Simulates rolling a 6-sided die.
```

samples/excel/16-custom-functions/volatile-function.yaml
davidchesnut: I'd love to change this, but yeah, I'm worried many folks may think that is a cell reference. :-)

PR #289, opened 2019-04-05
Update formula for roll6sided dice

Zlatkovsky:
To make it a fair dice, you should just choose a random number 0 to 1, multiply by 6, take the FLOOR of it, and add 1.  I.e.: `Math.floor(Math.random() * 6) + 1;`

The previous formula, `Math.round(Math.random() * 5 + 1)`, does not produce the same distribution


Zlatkovsky:
@davidchesnut , please review and then merge in.

AlexJerabek:
@davidchesnut Merged this for you :-)

PR #294, opened 2019-04-19
[excel] (Comments) Make comment snippet more robust

AlexJerabek:
This snippet unintentionally required the user to have cell A1 selected. That has been corrected.


AlexJerabek:
Thanks @kbrandl. I fixed the edit comment target. I also clarified the experience more in the task pane.

PR #296, opened 2019-04-24
[excel] (PivotTable, Slicer) Adding a slicer snippet

AlexJerabek:
This snippet covers the [Slicer](https://docs.microsoft.com/en-us/javascript/api/excel/excel.slicer?view=office-js) APIs. It shows how to add, format, delete, and adjust the filters of a slicer on a PivotTable.


samples/excel/85-preview-apis/pivottable-slicer.yaml
```diff
@@ -0,0 +1,211 @@
...
+          <button id="setup" class="ms-Button">
+                <span class="ms-Button-label">Add data</span>
+            </button>
+          <p />
+          <button id="add-pivot-table" class="ms-Button">
+            <span class="ms-Button-label">Add pivot table</span>
+          </button>
+        </section>
+        <section class="samples ms-font-m">
+          <h3>Try it out</h3>
```
Rick-Kirkham: If there's an order in which the buttons in the Try it Out section should be pushed, might want to add an instruction about that.

samples/excel/85-preview-apis/pivottable-slicer.yaml
```diff
@@ -0,0 +1,211 @@
...
+        $("#format-slicer").click(() => tryCatch(formatSlicer));
+        $("#apply-style").click(() => tryCatch(applyStyle));
+        $("#add-filters").click(() => tryCatch(addFilters));
+        $("#remove-filters").click(() => tryCatch(removeFilters));
+        $("#remove-slicer").click(() => tryCatch(removeSlicer));
+
+        async function addSlicer() {
+            await Excel.run(async (context) => {
+                const sheet = context.workbook.worksheets.getItem("Pivot");
+                const slicer = sheet.slicers.add("Farm Sales", "Type");
```
Rick-Kirkham: This is going to be puzzling to beginners with the slice APIs. The first parameter of the slicers.add method looks like it is the name of slice, but it's actually the name of the source. A code comment would help here.

AlexJerabek:
Thanks @Rick-Kirkham. I've updated the PR based on your feedback.

As per the read-properties comment: I'm not sure that's a standard scenario for slicers. There's a standard formatting they have when created, so it's more valuable to show how to set those properties. I don't see anything on [the API](https://docs.microsoft.com/en-us/javascript/api/excel/excel.slicer?view=office-js) that's terribly interesting to read under normal circumstances.

PR #305, opened 2019-05-23
added basic custom function

davidchesnut:
This just shows how to write a basic custom function. In this case it calculates the volume of a sphere.

### DIFF
--- a/samples/excel/16-custom-functions/basic-function.yaml
+++ b/samples/excel/16-custom-functions/basic-function.yaml
@@ -1,0 +1,22 @@
+order: 1
+id: excel-custom-functions-basic
+name: Basic custom function
+description: A simple custom function that calculates the volume of a sphere.
+host: EXCEL
+api_set:
+    CustomFunctionsRuntime: 1.1
+script:
+    content: |
+        /**
+         * Calculates the volume of a sphere. 
+         * @customfunction
+         * @param {number} radius
+         */
+        function sphereVolume(radius) {
+          return Math.pow(radius, 3) * 4 * Math.PI / 3;
+        }
+    language: typescript
+libraries: |
+    https://appsforoffice.microsoft.com/lib/1/hosted/office.js
+    @types/office-js
+    core-js@2.4.1/client/core.min.js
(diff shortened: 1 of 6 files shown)


samples/excel/16-custom-functions/basic-function.yaml
```diff
@@ -0,0 +1,22 @@
+order: 1
+id: excel-custom-functions-basic
+name: Basic custom function
+description: A simple custom function that calculates the volume of a sphere.
```
AlexJerabek: In the other Excel snippets, we've had description naming patterns of "Verbs X." Like "Creates a few different geometric shapes and deletes them from the worksheet." I'm not sure that pattern makes sense for CFs, but something to consider.

davidchesnut:
@AlexJerabek this particular custom function doesn't show any APIs that would be in the reference, but that is a good recommendation to consider as we start adding additional ones, such as calling the dialog api.

samples/excel/16-custom-functions/basic-function.yaml
jakobpn: We could just say "Calculate the volume of a sphere" since the title already have "custom function" in it.

samples/excel/16-custom-functions/basic-function.yaml
```diff
@@ -0,0 +1,22 @@
...
+host: EXCEL
+api_set:
+    CustomFunctionsRuntime: 1.1
+script:
+    content: |
+        /**
+         * Calculates the volume of a sphere. 
+         * @customfunction
+         * @param {number} radius
+         */
```
jakobpn: Do you also want to show the result in the JS doc comments?

davidchesnut: Fixed!

samples/excel/16-custom-functions/basic-function.yaml
davidchesnut: Fixed!

davidchesnut:
@Zlatkovsky I think Alex is out for a couple days. If these changes look okay can you merge this PR? Thanks!

davidchesnut:
Thanks @AlexJerabek . I don't have perms to merge on this repo. Can you do the merge or does someone else do that?

davidchesnut:
@AlexJerabek okay never mind I guess I do have admin permissions. I'll go ahead and merge. Thanks!

PR #306, opened 2019-06-05
[admin] (Dependencies) Updating fstream version

AlexJerabek:
Addressing the security vulnerabilities in this repo.


Zlatkovsky:
Yeah, I tend to agree with @Rick-Kirkham , that based on my understanding, just modifying the yarn.lock file alone is probably not the right way.
Ideally you either upgrade to the latest-and-greatest of the top-level package, OR you can provide an override as follows:  https://yarnpkg.com/lang/en/docs/selective-version-resolutions/
Give it a try and let me know how it goes.  If you need my help, come find me.

AlexJerabek:
@Zlatkovsky If that's the case, should the "npm start" commands in the readme be replaced by yarn commands?

Zlatkovsky:
@AlexJerabek , It should be one or the other.  `yarn` is generally considered better (other than requiring having `yarn` in the first place, but you can always obtain it with `npm install -g yarn` -- so I guess that should be added to the readme too).
Or stick with just plain NPM.  But in that case delete the yarn lock file and add it to the gitignore.

AlexJerabek:
@Zlatkovsky I adjusted the readme to be consistent about yarn. Also, running yarn install and yarn start didn't seem to generate a yarn.lock file. Is that expected?

PR #312, opened 2019-06-18
[all] (office-js-helpers) Removing references to the office-js-helpers library

AlexJerabek:
The biggest changes were removing the calls to forceCreateSheet in nearly every Excel sample. Otherwise, it was mostly library reference removals.


samples/excel/10-chart/chart-axis-formatting.yaml
```diff
@@ -52,8 +52,10 @@ script:
 
         async function setup() {
             await Excel.run(async (context) => {
-                let sheet = await OfficeHelpers.ExcelUtilities
-                    .forceCreateSheet(context.workbook, "Sample");
+                context.workbook.worksheets.getItemOrNullObject("Sample").delete();
```
Zlatkovsky: With here and anywhere you're removing "forceCreateSheet":  are you sure that the sample would only have worked on ExcelApi 1.4 and higher?

If so, great.  But I thought some samples (e.g., report generation) were supposed to work on older (RTM) versions of Office 2016 as well.  Whereas by you making use of the "OrNullObject" functionality, you are effectively breaking that.

See https://github.com/OfficeDev/office-js-helpers/blob/master/src/excel/utilities.ts#L103

samples/excel/44-shape/shape-textboxes.yaml
```diff
@@ -72,7 +72,10 @@ script:
 
         async function setup() {
             await Excel.run(async (context) => {
-                await OfficeHelpers.ExcelUtilities.forceCreateSheet(context.workbook, "Shapes");
+                context.workbook.worksheets.getItemOrNullObject("Shapes").delete();
+                const sheet = context.workbook.worksheets.add("Shapes");
+                await context.sync();
```
Zlatkovsky: Also, here and elsewhere, you probably don't need the "context.sync()" here (it's only causing an extra round-trip for absolutely no good reason).  The reason the original method was awaitable was because it *may* have needed to do a sync (but it only did it if it had to on older 2016 clients).  Whereas here you are doing it all the time.

samples/excel/46-table/sort-data.yaml
```diff
@@ -13,56 +13,55 @@ script:
...
-                {
-                key: 3,
-                ascending: false
-                }
-            ];
-            expensesTable.sort.apply(sortFields);
-
-            await context.sync();
+                const sheet = context.workbook.worksheets.getItem("Sample");
+                const expensesTable = sheet.tables.getItem("ExpensesTable");
```
Zlatkovsky: Were these just always badly indented and you're now fixing them?

AlexJerabek: Yeah, figured I might as well clean this up while I'm in there.

samples/excel/10-chart/chart-axis-formatting.yaml
AlexJerabek: Office 2016 only supports ExcelApi 1.1, so most of the samples don't work there anyway. Since everything else should be relatively up to date, I'm not sure there's value in targeting 1.2-1.4 era builds.

Plus, we've never called out required versions in SL snippets. Perhaps that's a valid thing to do in the future, but I feel like it's out of scope here.

samples/excel/44-shape/shape-textboxes.yaml
AlexJerabek: Before I go back through and test every Excel snippet to ensure this `context.sync()` is not necesssary, isn't it valid because of the delete beforehand? It ensures the workbook is in a ready state for what follows.

samples/excel/10-chart/chart-axis-formatting.yaml
Zlatkovsky: @AlexJerabek , each sample has an API set at the top of the file in the YAML, which Script Lab does use.  E.g,.:

```
api_set:
    ExcelAPI: 1.8
```

So, what I'm saying is that for any sample where you've removed "forceCreateSheet", you should at minimum make sure that the API set of that sample was >= 1.4, and/or update it to be greater than 1.4.

When you do, for the report generation sample, you can also get rid of some of the other "isSetSupported" checks within it, since I believe it checks for 1.3 req set -- which won't be relevant once you make the whole sample require 1.4

Zlatkovsky: BTW, I think it would also be interesting (and easy?) to look at the OMEX telemetry to see the distribution of build numbers for people using Script Lab.  I.e., is it true that 90% (or whatever) of SL users are on reasonably recent builds?  Or is there still a non-trivial contingent of people who are on 2016 MSI?..

samples/excel/44-shape/shape-textboxes.yaml
Zlatkovsky: It's just not necessary, making the code slower and adding confusion for no good reason.
For the latter, what I mean is that it calls into question what "context.sync()" even means.  Sure, performance aside, it doesn't hurt to stick it after _every_ API call -- but that defeats the point of batching (and makes perf be terrible).  So having extra syncs for no good reason hurts the conceptual integrity of explaining when you need syncs (which should be "as rarely as possible, only when needed")

AlexJerabek:
@Zlatkovsky I've removed the extra `context.sync()` calls. I've adjusted the relevant API set requirements to 1.4 (though some like Report Generation didn't need to be changed). I filed a work item to go back and rewrite a couple samples to only use 1.1 APIs ([internal link](https://office.visualstudio.com/OC/_workitems/edit/3437897)).

PR #313, opened 2019-06-19
Added web call custom function

davidchesnut:
This adds a new custom function that shows how to call a web api. The web api is to GitHub to get the star count for a specified repo.


samples/excel/16-custom-functions/web-call-function.yaml
```diff
@@ -0,0 +1,38 @@
...
+            const response = await fetch(url);
+
+            //Expect that status code is in 200-299 range
+            if (!response.ok) {
+              throw new Error(response.statusText);
+            }
+            const jsonResponse = await response.json();
+
+            return jsonResponse.watchers_count;
+          } catch (error) {
```
jakobpn: I don't think you need to have a try-catch that does nothing with the error expect throwing it again.

samples/excel/16-custom-functions/web-call-function.yaml
```diff
@@ -0,0 +1,38 @@
...
+         * Gets the star count for a given org/user and repo. Try =GETSTARCOUNT("officedev","office-js")
+         * @customfunction
+         * @param userName string name of org or user
+         * @param repoName string name of the repo
+         * @return number of stars.
+         */
+        async function getStarCount(userName: string, repoName: string) {
+          try {
+            //You can change this URL to any web request you want to work with.
+            const url = "https://api.github.com/repos/" + userName + "/" + repoName;
```
jakobpn: Since it is TypeScript you can also use a template:   const url = `https://api.github.com/repos/${userName}/${repoName}`;
It is a little shorter and less " and + to keep track of.

samples/excel/16-custom-functions/web-call-function.yaml
```diff
@@ -0,0 +1,38 @@
...
+            const response = await fetch(url);
+
+            //Expect that status code is in 200-299 range
+            if (!response.ok) {
+              throw new Error(response.statusText);
+            }
+            const jsonResponse = await response.json();
```
jakobpn: I suggest keeping this line with the next line instead of the previous block. So insert blank line before rather than after.

jakobpn:
Very nice to have this snippet added. If there are more properties on a GitHub repo we should consider adding a sample that show how to cache the result of a web request for performance.

samples/excel/16-custom-functions/web-call-function.yaml
davidchesnut: @jakobpn excellent suggestion! Thank you!

samples/excel/16-custom-functions/web-call-function.yaml
davidchesnut: fixed.

samples/excel/16-custom-functions/web-call-function.yaml
davidchesnut: fixed.

AlexJerabek:
@jakobpn, Were your concerned addressed by David's revisions?

PR #315, opened 2019-06-21
[excel] (Samples) Mapping snippets for structs that must be set together

AlexJerabek:
This maps some snippets to APIs that are difficult to use. The properties in question must be assigned as JSON objects, not through the setting of individual subproperties. By surfacing more sample code, we can mitigate the customer impact.

This PR also adds a `PageLayout.zoom` sample.


snippet-extractor-output/snippets.yaml
```diff
@@ -1874,6 +1937,19 @@ Excel.Range.style:
...
+        await context.sync();
+    });
+Excel.RangeAreas.format:
+  - |-
+    await Excel.run(async (context) => {
+
+        const sheet = context.workbook.worksheets.getActiveWorksheet();
+        const usedRange = sheet.getUsedRange();
+
+        // Find the ranges with formulas
```
ElizabethSamuel-MSFT:
```suggestion
        // Find the ranges with formulas.
```

Issue #318, opened 2019-06-26
Modify "gradients" sample to make use of setCellProperties


Zlatkovsky:
I've seen two questions on StackOverflow about getting or setting cell properties in bulk, the most recent of which was yesterday:  https://stackoverflow.com/questions/56755657/how-does-the-range-function-getcellproperties-in-the-javascript-excel-api-work.

The “Just for Fun -> Gradient” sample would be trivially convertible to make use of the setCellProperties API, and would run an order of magnitude faster than it does today (or faster still on larger sizes).

You could even make it be two different helper functions, that are redirected to based on whether `isSetSupported("ExcelApi", 1.9)` is true.

I'm happy to review the PR.
- Michael

AlexJerabek:
Thanks @Zlatkovsky. We'll look into improving that sample and writing a separate one that'll be more searchable. [Internal task link](https://office.visualstudio.com/OC/_workitems/edit/3455984).

AlexJerabek:
Fixed with #320. Closing this issue.

PR #320, opened 2019-07-02
[excel] (Range) Upgrading Gradient snippet to showcase setCellProperties API

AlexJerabek:
Based on the suggestion from #318.


samples/excel/99-just-for-fun/gradient.yaml
```diff
@@ -41,29 +40,68 @@ script:
...
-            fillInColors(colors2D, originalSize);
-
-            let range = sheet.getCell(1, 1).untrack().getResizedRange(
-                originalSize - 1, originalSize - 1).untrack();
-            for (let row = 0; row < originalSize; row++) {
-                for (let col = 0; col < originalSize; col++) {
-                    range.getCell(row, col).untrack().format.fill.color =
-                        colors2D[row][col];
+            if (Office.context.requirements.isSetSupported("ExcelApi", 1.9)) {
+                // ExcelApi 1.9 introduced the setCellProperties APIs to quickly set different properties across a range.
```
Zlatkovsky: You might want to expand on the "quickly" a bit.  E.g.,
> // ExcelApi 1.9 introduced the setCellProperties APIs to **efficiently** set different properties
> across a range **without needing to iterate cell-by-cell**.

samples/excel/99-just-for-fun/gradient.yaml
```diff
@@ -41,29 +40,68 @@ script:
...
+                    }
                 }
             }
 
             sheet.activate();
             await context.sync();
         }
 
+        function fillInColorsWithCellProperties(colors2D: Excel.SettableCellProperties[][], size: number) {
+            const topColors = getColorsArray(
```
Zlatkovsky: I think the code would be much cleaner if you just re-use the `fillInColors` function and then `.map` over it to create the data structure you want.  Otherwise, you are copying the code even though 90% of it could have been re-usable...

Also, I think the creation of the `colors2D` array of the original could code really move into this function (making it a `create2DColorsArray` function).  That would further clean up and consolidate the logic.

AlexJerabek:
@Zlatkovsky Feedback addressed. Was this the code consolidation you were imagining?

Zlatkovsky:
@AlexJerabek , it's much closer to what I was imagining, but I think it could sill be improved.

Right now, your helper function returns different object types for whether it's ExcelApi 1.9 or not.  And worse, it mixes in "Excel-yness" into logic that feels like it should be totally independent of Excel (you could imagine a developer wanting to write code like this for painting the HTML dom elements; and also wanting to unit-test it, etc, without obeying by the "proprietary" format of the setCellProperties API.

So I think what I'd suggest is instead having `createColorsArray` always return a `string[][]` (i.e., the logic of the "else" part at the end of the function).  And then in the part where you're consuming it, do:

```
const cellProperties: Excel.SettableCellProperties[][] =
	colors2D.map(row =>
		row.map(color =>
			({
				format: {
					fill: {
						color: color
					}
				}
			})
		)
	);
sheet.getRangeByIndexes(1, 1, originalSize, originalSize).setCellProperties(cellProperties);
```

PR #321, opened 2019-07-03
[typo] (Snippet Mapping) Remove spaces from mapping file

AlexJerabek:
Five spaces wound up in the Excel file. Removing them so the classes map properly.


AlexJerabek:
After this builds, I'll check it in, since there's not really a way to review this file.

PR #352, opened 2019-09-18
[excel] (Range) Group/Ungroup sample

AlexJerabek:
This sample shows how to use the outline (group/ungroup) functionality with ranges. It creates a two-level outline on both dimensions.


AlexJerabek:
@kbrandl That is the exact same experience Excel delivers. With the rows and columns ungrouped, the user is left with tiny rows and columns where the groups were (not technically, but practically hidden).

kbrandl:
@AlexJerabek fair enough! Thanks for confirming :)

PR #354, opened 2019-09-20
[excel] (Range) Auto fill sample

AlexJerabek:
This sample shows how to use `Range.autoFill`. It was a request from the ECO team.


samples/excel/85-preview-apis/range-auto-fill.yaml
```diff
@@ -0,0 +1,156 @@
...
+          <button id="setup-data" class="ms-Button">
+            <span class="ms-Button-label">Add sample data</span>
+        </button>
+          <p />
+          <button id="setup-first-sums" class="ms-Button">
+            <span class="ms-Button-label">Add first sum cells</span>
+        </button>
+        </section>
+
+        <section class="samples ms-font-m">
```
kbrandl: Recommend shortening (rewording) each of these 3 button labels, if possible. Maybe something along these lines instead?
- `Auto fill values in column F`
- `Auto fill formats in column K`
- `Auto fill everything in column P`

As buttons are currently labeled, the length of each button label is too long to fit within the button, unless I resize my "Run" pane to be very wide (which'll hinder usability of the sample for folks using smaller screens -- since the "Run" pane having to be so wide will make them have to manually scroll side to side to see all of the data in the worksheet).

image

AlexJerabek:
Good call. Relabeled the buttons with your suggestions.

PR #363, opened 2019-10-22
[excel] (ExcelApi 1.10) Moving 1.10 snippets out of preview

AlexJerabek:


samples/excel/42-range/range-auto-fill.yaml
```diff
@@ -143,8 +143,8 @@ style:
         }
     language: css
 libraries: |
-    https://appsforoffice.microsoft.com/lib/beta/hosted/office.js
-    @types/office-js-preview
+    https://appsforoffice.microsoft.com/lib/1/hosted/office.js
+    @types/office-js
```
kbrandl: Does the (sort) `order` attribute value need to be updated for this snippet (like you've done for other snippets being moved out of the preview folder into another folder)?

AlexJerabek: This snippet was conveniently already in the right order. Thanks for looking out though.

PR #381, opened 2019-11-25
[Outlook] Adds attachment samples

Oleg-O:


samples/outlook/40-attachments/get-attachment-content.yaml
```diff
@@ -0,0 +1,87 @@
...
+                        // Handle attachment formats that are not supported.
+                }
+            }
+        }
+    language: typescript
+template:
+    content: |
+        <section class="ms-font-m">
+          <p>This sample shows how to get the attachment content from a message or an appointment.</p>
+          <p><b>Required mode</b>: Item Compose</p>
```
ElizabethSamuel-MSFT: This should work in Compose or Read mode, right?

PR #387, opened 2019-12-19
[excel] (Range) Adding a Range.moveTo example

AlexJerabek:


samples/excel/42-range/range-copyfrom.yaml
```diff
@@ -110,12 +111,23 @@ script:
             });
         }
 
+        async function move() {
```
ElizabethSamuel-MSFT: Should this be a standalone sample? The name of this is copyfrom so I don't know that it'll be clear that move is included here. Maybe rename this to something like copyAndMove?

samples/excel/42-range/range-copyfrom.yaml
```diff
@@ -110,12 +111,23 @@ script:
...
+            await Excel.run(async (context) => {
+                const sheet = context.workbook.worksheets.getItem("Sample");
+                sheet.getRange("F12").values = [["Moved Range"]];
+                sheet.getRange("F12").format.autofitColumns();
+
+                // move a range from a range to the specified location
```
ElizabethSamuel-MSFT:
```suggestion
                // Move a range to the specified location.
```

samples/excel/42-range/range-copyfrom.yaml
AlexJerabek: I thought about separating it out, but I think it's better off here. It's part of the same "cut/copy/paste" workflow. I'll add the word "move" to the description to help with searches.

PR #400, opened 2020-02-24
Excel scenarios currency converter

cakriwut:
**Description**  
Demonstrate calling currency converter API and calculate the equivalent value in local currency.
Demonstrate dynamic address range finder based on table specification.

**Target Host**
Excel


samples/excel/90-scenarios/currency-converter.yaml
```diff
@@ -0,0 +1,178 @@
+order: 5
+id: excel-scenarios-currency-converter
+name: Currency Converter
+description: 'Reads rows and update base currency column, based on value date.'
```
AlexJerabek: Let's change the description to better highlight the external API being used.
"Uses an exchange rate API to convert currency values based on their original transaction times."

samples/excel/90-scenarios/currency-converter.yaml
```diff
@@ -0,0 +1,178 @@
...
+        }
+      }
+      /* Create sample data */
+      /* Default header */
+      /* "CCY","Price (CCY)","Trx Date","Base","Price in Base" */
+      async function setupSample() {
+        await Excel.run(async (context) => {
+          context.workbook.worksheets.getItemOrNullObject("Sample").delete();
+          const sheet = context.workbook.worksheets.add("Sample");
+          let salesTable = sheet.tables.add("A1:F1", true);
```
AlexJerabek: The table is less about sales and more of a transaction history. Maybe call it `transactionTable`

samples/excel/90-scenarios/currency-converter.yaml
```diff
@@ -0,0 +1,178 @@
...
+
+      /** Main converting function **/
+      async function convert() {
+        await Excel.run(async (context) => {
+          let range = context.workbook.getSelectedRange(); 
+          // Check if any cell inside table is selected.
+          let table = range.getTables();
+          table.load("items")    
+          await context.sync();
+          if(table.items.length == 0){
```
AlexJerabek: Since we name the table in `setupSample`, we can use `TableCollection.getItem` to reference it. That way, we don't need all this error checking. We also don't need to have the user select the table.

samples/excel/90-scenarios/currency-converter.yaml
```diff
@@ -0,0 +1,178 @@
...
+              if (priceInBase === "") {
+                let dateMoment = moment.fromOADate(trxDate);
+                let period = dateMoment.format("YYYY-MM-DD");
+                let queryUrl = convertValue(ccy, baseCcy, period, period);
+                let response = await fetch(queryUrl);
+                if (response.ok) {
+                  let result = await response.json();
+                  let converted = priceInCcy * result.rates[baseCcy];
+                  priceBaseRange.values = [[converted]];
+                  context.sync();
```
AlexJerabek: You need to add an `await` to this `context.sync()`. Otherwise there's a race condition (try your sample with 3+ rows).

samples/excel/90-scenarios/currency-converter.yaml
```diff
@@ -0,0 +1,178 @@
...
+        <section class="setup ms-font-m">
+            <h3>Set up</h3>
+            <button id="setup" class="ms-Button">
+                <span class="ms-Button-label">Add sample data</span>
+            </button>
+        </section>
+
+        <section class="samples ms-font-m">
+            <h3>Try it out</h3>
+            <p>The price conversion will be provide by conversion API. Select on empty price cell in price column, and click "Convert"</p>
```
AlexJerabek: The snippet runs on every rows in the transaction table, so the user doesn't need to select an individual cell. Instead, can you change this line to highlight the API being used? Like, `<p>The currency conversion is provided by <a href="https://exchangeratesapi.io">exchangeratesapi.io</a> which uses exchange rate data published by the European Central Bank.</p>`

samples/excel/90-scenarios/currency-converter.yaml
```diff
@@ -0,0 +1,178 @@
...
+            <ol>
+              <li>Select any cell in the table.</li>
```
AlexJerabek: If we remove the need to select the table, we don't need these steps.

samples/excel/90-scenarios/currency-converter.yaml
```diff
@@ -0,0 +1,178 @@
...
+              } else {
+                console.info('Row: ' + idx + ' has been converted. Skipped.')
+              }
+          });
+          
+        });
+      }
+
+      /** Check if Table is Valid, ie has all required column **/
+      function IsTableValid(items: Excel.TableColumn[]) {
```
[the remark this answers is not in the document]
cakriwut: IsTableValid allows user to re-arrange or create their own table. As long as the table has 5 pre-defined column header in their custom table. Unlike other samples that only show how to use static table - in this example I also demo how to handle case with unknown column position.
@AlexJerabek your final suggestion please.

package-lock.json
```diff
@@ -0,0 +1,733 @@
+{
```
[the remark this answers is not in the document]
cakriwut: Updated

samples/excel/90-scenarios/currency-converter.yaml
```diff
@@ -0,0 +1,178 @@
...
+          context.workbook.worksheets.getItemOrNullObject("Sample").delete();
+          const sheet = context.workbook.worksheets.add("Sample");
+          let salesTable = sheet.tables.add("A1:F1", true);
+          salesTable.name = "SalesTable";
+          salesTable.getHeaderRowRange().values = [["Product", "CCY", "Price (CCY)", "Trx Date", "Base", "Price in Base"]];
+          salesTable.rows.add(null, [
+            ["Frames", "MYR", 5000, "2019-10-05", "SGD", null],
+            ["Chains", "CNY", 12000, "2019-10-04", "SGD",null]
+          ]);
+          sheet.activate();
```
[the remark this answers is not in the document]
cakriwut: Updated

samples/excel/90-scenarios/currency-converter.yaml
```diff
@@ -0,0 +1,178 @@
...
+      /* Create sample data */
+      /* Default header */
+      /* "CCY","Price (CCY)","Trx Date","Base","Price in Base" */
+      async function setupSample() {
+        await Excel.run(async (context) => {
+          context.workbook.worksheets.getItemOrNullObject("Sample").delete();
+          const sheet = context.workbook.worksheets.add("Sample");
+          let salesTable = sheet.tables.add("A1:F1", true);
+          salesTable.name = "SalesTable";
+          salesTable.getHeaderRowRange().values = [["Product", "CCY", "Price (CCY)", "Trx Date", "Base", "Price in Base"]];
```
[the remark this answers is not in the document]
cakriwut: Updated, include change CCY => Currency

samples/excel/90-scenarios/currency-converter.yaml
cakriwut: Updated.

samples/excel/90-scenarios/currency-converter.yaml
```diff
@@ -0,0 +1,178 @@
...
+      declare let moment: any;
+      let tblSetup: Record<string, number> = {};
+
+      $("#convert").click(() => tryCatch(convert));
+      $("#setup").click(() => tryCatch(setupSample));      
+
+      /** Main converting function **/
+      async function convert() {
+        await Excel.run(async (context) => {
+          let range = context.workbook.getSelectedRange();
```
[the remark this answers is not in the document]
cakriwut: Updated.

samples/excel/90-scenarios/currency-converter.yaml
cakriwut: Updated.

samples/excel/90-scenarios/currency-converter.yaml
cakriwut: Removed

samples/excel/90-scenarios/currency-converter.yaml
cakriwut: Updated.

samples/excel/90-scenarios/currency-converter.yaml
cakriwut: Deleted.

samples/excel/90-scenarios/currency-converter.yaml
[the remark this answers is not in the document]
cakriwut: @AlexJerabek I keep this function. Added description line in HTML, that user can rearrange the column.

samples/excel/90-scenarios/currency-converter.yaml
```diff
@@ -0,0 +1,178 @@
...
+
+        /** Async ForEeach */
+        async function asyncForEach(array: Array<any>, callback) {
+          for(let idx = 0; idx < array.length; idx++) {
+            await callback(array[idx],idx, array);
+          }
+        }
+
+        /** Check if Table is Valid, ie has all required column **/
+        function IsTableValid(items: Excel.TableColumn[]) {
```
AlexJerabek: Thanks for clarifying the purpose of this method. Since you want to highlight the possibility of a dynamic table ordering, let's make that clear in the comment.
"Check if the table contains the necessary columns. These columns can be in any order within the table."

samples/excel/90-scenarios/currency-converter.yaml
```diff
@@ -0,0 +1,178 @@
...
+                console.info(`Row: ${idx} has been converted. Skipped.`)
+              }
+            });
+          });
+        }
+
+        /** Async ForEeach */
+        async function asyncForEach(array: Array<any>, callback) {
+          for(let idx = 0; idx < array.length; idx++) {
+            await callback(array[idx],idx, array);
```
AlexJerabek: The function name is misleading, since the for each isn't asynchronous. Please correct me if I'm wrong, but the `await` makes each callback complete before proceeding. I'm not sure you need all this either. Wasn't the original issue that we just needed to await a `context.sync()`?

samples/excel/90-scenarios/currency-converter.yaml
```diff
@@ -0,0 +1,178 @@
...
+script:
+    content: |
+        /*
+        * Copyright (c) Riwut Libinuko. All rights reserved. Licensed under the MIT license.
+        */
+
+        declare let moment: any;
+        const tableName = "TransactionTable";
+
+        let tblSetup: Record<string, number> = {};
```
AlexJerabek: Nit: Can we rename `tblSetup` to `tableSetup`?

samples/excel/90-scenarios/currency-converter.yaml
cakriwut: Updated.

samples/excel/90-scenarios/currency-converter.yaml
cakriwut: The original Array.ForEach(async () => {}) , does not await context.sync(). I tested using following test step.
1. Add 20 rows
2. Click "Convert"  ==> OK
3. Delete the converted value (Price In Base)
4. Click "Convert" ==> ERROR

The error - 
--
message: "The object path 'tables.getItem().rows.getItemAt().getRange().getCell()' isn't working for what you're trying to do. If you're using the object across multiple "context.sync" calls and outside the sequential execution of a ".run" batch, please use the "context.trackedObjects.add()" and "context.trackedObjects.remove()" methods to manage the object's lifetime."
---

samples/excel/90-scenarios/currency-converter.yaml
cakriwut: Updated.

samples/excel/90-scenarios/currency-converter.yaml
AlexJerabek: @cakriwut  Do you need to track the objects? All of this is in the same `Excel.run` call and uses the same context object. Check out this gist that removes the tracking and async for loop: https://gist.github.com/AlexJerabek/600b32fbc2787a13367d8b65ec133f16

I tested it with 120+ rows and never hit a race condition.

cakriwut: @AlexJerabek , it is strange that you didn't face the issue, in my Excel (Microsoft Office Professional Plus 2019, version 1911 build 12228.20250) - I have the issue in the next run.
My steps
1. Create 50 samples row
2. Click Convert  (I found no issue in this step)
3. Clear "Price In Base"
4. Click Convert (the race condition appears).

See the video here,  - https://drive.google.com/file/d/1QDCNrS39XDiixrRegtAFgQhNKMCPKSvX/view?usp=sharing

cakriwut: @AlexJerabek , I replace using for( let item of Array) instead. This solve the await inside Array.forEach

Rick-Kirkham:
@cakriwut @AlexJerabek I've been working on this submission. 

**Some discoveries (which you two may already have known):**

Using Alex's gist, I got the same error as cakriwut got in his video, even with just 2 rows in the table: The first row calculates correctly, then an error that suggests using `trackedObjects`. Adding cakriwut's `trackedObjects.add` line does seem to fix it for a small number of rows, but with a large number (64) I start getting the race condition that cakriwut was getting. However, Alex is correct that the `trackedObjects` isn't logically necessary. In fact, I tried tracking the `row` object instead of the `priceBaseRange` object and got the same relative benefit. Similarly, tracking the `table` object gives the same relative benefit. I *think* (very speculative) that the reason it seems to improve things (when there's a small number of rows) is simply because it slows down each iteration of the loop, and that the amount of the slow down is slightly variant between iterations (which would not be the case if you added a `setTimeout()` for a given number of seconds instead of a `trackedObjects.add`). 

The source of the race condition seems to be that the function passed to `forEach` is marked as `async`. This means that all of the iterations are fired off in quick succession without waiting for any preceding iteration to complete. You can see the cells in the **Price in Base** column filling randomly and very rapidly (until you get the Exception caused by the race condition). The race condition is one side effect of running the iterations asynchronously. Another is that Office thinks it's in a scenario where `trackedObjects` needs to be used. 

When I replace `forEach` loop with a `for` loop, so the iterations are synchronous, everything works fine. No race conditions, and you can see the cells in the **Price in Base** column filling in order from top to bottom, row-by-row; but much slower than with the asynchronous `forEach` loop. I can also remove all the `trackedObjects` code and it still doesn't break it. (@cakriwut, you mentioned in our last comment that you also switched to a `for` loop. Did that fix it for you too?)

**Recommendations:**

1. For performance reasons, it is not a good practice to have a `context.sync` inside a loop because each one is a separate round-trip between Office and the document, which can be significant, especially in Office on the web (aka: Office Online). If possible, I think the Script Lab snippets should show the best practices. There is a pattern that I call "Correlated Objects pattern" for avoiding loops with `context.sync` in them. Please see my answers to the following Stack Overflow questions for details:

[iterating-over-all-paragraphs-with-content-controls-with-office-javascript-api](https://stackoverflow.com/questions/58422113/iterating-over-all-paragraphs-with-content-controls-with-office-javascript-api)

[word-addin-inserting-complex-list-structure-with-styling-in-a-contentcontrol-usi](https://stackoverflow.com/questions/55191316/word-addin-inserting-complex-list-structure-with-styling-in-a-contentcontrol-usi)

[document-not-in-sync-after-replace-text](https://stackoverflow.com/questions/48227941/document-not-in-sync-after-replace-text)

2. There's an `await` before the `fetch` method. This turns the asynchronous method into a synchronous one and cancels the performance advantage that asynchronous methods give you. Sometimes, of course, you have to await it. That's why `await` was invented. But this is a classic case in which there's no reason why the `fetch`'s can't be running asynchronously: The data returned by each `fetch` is not dependent in any way on the data returned by a previous call of `fetch`. We should be able to remove that `await` and add each `response` object that's returned to an array. Then use [Promise.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) to await the array. It won't be quite that simple because the fetch calls won't necessarily complete in the order in which they run, so there needs to be a way of mapping each response with it's row number. Maybe something like this (pseudo code, not tested!!!!):

```
const unMappedRowsData = []; // Created only so Promise.all can be called on it.
const mappedRowsData = [];

// in a loop
var response = fetch(queryUrl); // Don't await!
unMappedRowsData.push(response);

// Map the response to the row that it goes with.
var mappedRow = { rowNumber: row.index, 
                  data: response };
mappedRowsData.push(mappedRow);
// end loop

Promise.all(unMappedRowsData );
// At this point all the response objects have been resolved, so 
// we can now iterate on the mappedRowsData array, call `json()` 
// on the `data` property of each (that is, on the `response` object),
// and write the parsed data to its corresponding row.
```

The `json()` is also async, but it's just parsing the `response` object that is already in memory and that JSON blob is very small, so I don't think there's any performance concern with awaiting it. But, if we wanted to, we could use the same technique with Promise.all to parse all the response objects concurrently.

cakriwut:
Thanks @Rick-Kirkham for the insight, I have updated the code and tested with 1k rows without issue.

1.  Use Array.Map and return all converted value, wrapped in Promise.all. 
2. Apply the converted value to the affected rows

This approach improves the conversion time 10x faster.

Rick-Kirkham:
@cakriwut This is great! There's no `context.sync` in a loop. In fact, there's only one explicit `context.sync` in the whole `Excel.run`. The reason that the queue of `priceBaseRange.values = [[convertedCurrency[index]]];` commands syncs with the document, is because, under the hood, `Excel.run` checks to see if there are any unsynced commands in the queue just before it ends. If there are, it quietly calls `context.sync` to sync them. We usually consider it a best practice to have an explicit `constext.sync` as the last line of the `Excel.run` just before it's closing "}". The reason is that this makes it clear to someone reading the code that the sync has to happen. There's no danger of getting a pointless second `context.sync` because `Excel.run` will call the implicit `context.sync` only if there are unsynced commands in the queue and there won't be if the last line is an explicit `context.sync`. I'll leave it to @AlexJerabek to decide whether to enforce this best practice in this sample (but I would vote in favor of adding the explicit `context.sync`).

I notice that the latest version is still awaiting the `fetch`. Since this is not an API in Office.js, I'm not concerned about that; but you should consider someday making a version that runs all those fetches concurrently as I suggested earlier.

cakriwut:
@Rick-Kirkham I have removed await fetch, adding context.sync()

Rick-Kirkham:
Looks great! Thanks.

PR #405, opened 2020-03-12
[excel] (Calculation APIs) Workbook calculation snippet

AlexJerabek:
This snippet was originally designed to highlight the new `WorksheetCalculatedEventArgs.address` property, but seemed like a good place to showcase more related APIs.

### DIFF
--- a/playlists/excel.yaml
+++ b/playlists/excel.yaml
@@ -880,7 +880,7 @@
   fileName: workbook-calculation.yaml
   description: >-
     Demonstrates the calculation APIs of the workbook: worksheet-level
-    calulcation events and application-level calculation controls.
+    calculation events and application-level calculation controls.
   rawUrl: >-
     https://raw.githubusercontent.com/<ACCOUNT>/<REPO>/<BRANCH>/samples/excel/85-preview-apis/workbook-calculation.yaml
   group: Preview APIs
--- a/playlists/excel.yaml
+++ b/playlists/excel.yaml
@@ -875,6 +875,17 @@
   group: Worksheet
   api_set:
     ExcelApi: '1.1'
+- id: excel-workbook-calculation
+  name: Calculations
+  fileName: workbook-calculation.yaml
+  description: >-
+    Demonstrates the calculation APIs of the workbook: worksheet-level
+    calulcation events and application-level calculation controls.
+  rawUrl: >-
+    https://raw.githubusercontent.com/<ACCOUNT>/<REPO>/<BRANCH>/samples/excel/85-preview-apis/workbook-calculation.yaml
+  group: Preview APIs
+  api_set:
+    ExcelApi: '1.11'
 - id: excel-comment-mentions
   name: Comment mentions
   fileName: comment-mentions.yaml
--- a/samples/excel/85-preview-apis/workbook-calculation.yaml
+++ b/samples/excel/85-preview-apis/workbook-calculation.yaml
@@ -1,7 +1,7 @@
 order: 1
 id: excel-workbook-calculation
 name: Calculations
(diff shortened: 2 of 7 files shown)


samples/excel/85-preview-apis/workbook-calculation.yaml
```diff
@@ -0,0 +1,180 @@
+order: 1
+id: excel-workbook-calculation
+name: Calculations
+description: 'Demonstrates the calculation APIs of the workbook: worksheet-level calulcation events and application-level calculation controls.'
```
Rick-Kirkham: spell error "calulcation"

playlists/excel.yaml
```diff
@@ -875,6 +875,17 @@
   group: Worksheet
   api_set:
     ExcelApi: '1.1'
+- id: excel-workbook-calculation
+  name: Calculations
+  fileName: workbook-calculation.yaml
+  description: >-
+    Demonstrates the calculation APIs of the workbook: worksheet-level
+    calulcation events and application-level calculation controls.
```
Rick-Kirkham: spell "calulcation"

samples/excel/85-preview-apis/workbook-calculation.yaml
```diff
@@ -0,0 +1,180 @@
...
+          await Excel.run(async (context) => {
+            let sheet = context.workbook.worksheets.getItem("Sample");
+            sheet.onCalculated.add(onCalculated);
+            await context.sync();
+            console.log("Added worksheet selection changed event handler.");
+          });
+        }
+        async function onCalculated(event: Excel.WorksheetCalculatedEventArgs) {
+          await Excel.run(async (context) => {
+            console.log(`The Range ${event.address} has recalculated.`);
```
Rick-Kirkham: Could we have a comment that tells what **event.address** is?

samples/excel/85-preview-apis/workbook-calculation.yaml
```diff
@@ -0,0 +1,180 @@
...
+            const randomResult = Math.floor(Math.random() * 10000);
+
+            const cellToChange = sheet.getRange("B2:E7").getCell(randomRow, randomColumn);
+
+            cellToChange.load("address");
+            await context.sync();
+            console.log(`Changing cell ${cellToChange.address}`);
+
+            cellToChange.values = [[randomResult]];
+          });
```
Rick-Kirkham: Have we stopped adding an explicit final `context.sync` to the end of `Excel.run`? We always used to have that as a best practice even though it isn't necessary, because under the hood Excel.run will call context.sync if there are any unsynced commands in the queue.

samples/excel/85-preview-apis/workbook-calculation.yaml
```diff
@@ -0,0 +1,180 @@
...
+              ["Mirrors", 225, 600, 923, 544, "=SUM(B6:E6)"],
+              ["Spokes", 6005, 7634, 4589, 8765, "=SUM(B7:E7)"]
+            ]);
+
+            sheet.activate();
+            await context.sync();
+          });
+        }
+
+        /** Default helper for invoking an action and handling errors. */ async function tryCatch(callback) {
```
Rick-Kirkham: Seems to be a missing line break after the comment.

samples/excel/85-preview-apis/workbook-calculation.yaml
AlexJerabek: Good call. I added syncs to the end of those write methods. I left it out of methods that looked like this:
```
...
  await context.sync();
  console.log("Something");
});
```

Does that seem reasonable?

Rick-Kirkham: Yes.

PR #416, opened 2020-03-28
Gives user 3 different reading views for a excel sheet

PramodKumarYadav:
This snippet gives the user 3 reading views which are especially useful when reading big (wrapped text) excel sheets. These views are:
1. Balanced view (Which autofits the row shared by the current active cell thus giving user a full view of the row, keeping all other rows compressed). If user navigates to the next row, the other row expands keeping all other rows compressed.
2. Compressed view (which gives user a compressed view of the sheet. Useful when say user wants to see numbers or (X)s in columns and doesn't want the view to be distorted due to wrapped up text in other cells.
3. Expanded view (where each row is fully expanded. Could be useful for smaller sheets where user wants to see all information expanded in the sheet).
P.S: I find the usability of views in the order I have explained them with balanced view being the most useful of them all.

### DIFF
--- a/playlists/excel.yaml
+++ b/playlists/excel.yaml
@@ -984,6 +984,15 @@
   group: Scenarios
   api_set:
     ExcelApi: '1.4'
+- id: reading-views
+  name: Reading views
+  fileName: reading-views.yaml
+  description: Three different reading views for reading excel sheets.
+  rawUrl: >-
+    https://raw.githubusercontent.com/<ACCOUNT>/<REPO>/<BRANCH>/samples/excel/95-use-cases/reading-views.yaml
+  group: Use Cases
+  api_set:
+    ExcelApi: '1.9'
 - id: excel-just-for-fun-patterns
   name: Colorful Patterns
   fileName: patterns.yaml
--- a/samples/excel/95-use-cases/reading-views.yaml
+++ b/samples/excel/95-use-cases/reading-views.yaml
@@ -1,0 +1,189 @@
+order: 1
+id: reading-views
+name: Reading views
+description: Three different reading views for reading excel sheets.
+host: EXCEL
+api_set:
+    ExcelApi: '1.9'
+script:
+    content: |
+        $("#balanced-rows").click(() => tryCatch(balancedRows));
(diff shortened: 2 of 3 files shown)


AlexJerabek:
Hi @PramodKumarYadav, 

Thank you for submitting this snippet. We appreciate the contributions from the community. A quick sample about 

I loaded up your snippet and have a couple questions before delving deeper into the code:

- There doesn't seem to be a way to turn off "Balanced Rows" view. You can press the other two buttons, but the row width changes the instant you select a new cell. I'd expect clicking "expand" or "collapse" would turn off the row balancing.

- Instead of using a timer, please use the `Worksheet.onSelectionChanged` event. That will have better performance and show a more reusable pattern.

- There seem to be some bugs that come up if you change the width and modes enough. Sometimes the expand/collapse behavior inverts. Sometimes you get "current row is out of reading range. Go to used rows and trigger this add in again for a balanced reading view!" when scrolling through. The following gif shows some of this unpredictable behavior:

ReadingView

Since these snippets are teaching tools we expect others to copy-and-paste, they need to be bug-free and super simple to use. Please address the issues, then I'll be happy to take a deeper look.

Thanks!

PramodKumarYadav:
Hello Alex,
Thank you for having a look and your suggestions. I will have a look at your review comments in detail. 
On a quick glance, 1) -> the 'balanced rows' can be turned off, once you go out of used range of rows. I should mention it explicitly in the balanced row addin comment. 
2) I will try with your suggestion to use events. I am new with Javascript and in fact these were my first attempts in a day or two of work, so feel free to suggest any more changes, when you have a look.
3). Sometimes you get "current row is out of reading range. { this is when the exit criteria for balanced rows is met. Once I add the remark, users will see this as a confirmation).
I will try to replicate the errors you got with expand and collapse. In my earlier trials, I did not find any bugs with them.

Rick-Kirkham:
@PramodKumarYadav Please export this to a gist and post a link to the gist as a reply to this comment. Thanks.

Rick-Kirkham:
@PramodKumarYadav I'll close this for now. If you want to pursue it, please provide the gist and add a comment asking that it be reopened. Thanks for your interest in Office Add-ins.

PR #424, opened 2020-03-31
[excel] (Events) onSelectionChanged sample

AlexJerabek:
This sample showcases the behavioral differences between different `onSelectionChanged` events.

It also updates the tslint package version to match that of the other add-ins/scripts repos.


samples/excel/30-events/selection-changed-events.yaml
```diff
@@ -0,0 +1,160 @@
...
+
+            // Add a selection changed event handler for the worksheet collection.
+            context.workbook.worksheets.onSelectionChanged.add(onWorksheetCollectionSelectionChange);
+
+            await context.sync();
+          });
+        }
+
+        async function onBindingSelectionChange(args: Excel.BindingSelectionChangedEventArgs) {
+          await Excel.run(async (context) => {
```
Rick-Kirkham: You've probably already done this, but just for sanity I'll ask, have you verified that the `async` keyword is needed here? There don't seem to be any internal calls to `await`. Same question applies to the next few handlers defined below. Then, again, maybe its a good idea to keep it anyway since in most real life `Excel.run`s it would be needed. So, this is just something to think about.

AlexJerabek: This is definitely something to think about. I left it to maintain the typical `Excel.run` pattern we use in the TypeScript samples. Like you mentioned, most `Excel.run`s need the `async`/`await`, so this ensures there won't be any copy/paste hiccups for developers building off of this sample.

Happy to chat further and do an audit of the other samples, if needed.

PR #457, opened 2020-05-21
[Outlook] Shows how to inline base64 image

ElizabethSamuel-MSFT:


samples/outlook/99-preview-apis/work-with-client-signatures.yaml
```diff
@@ -61,6 +62,31 @@ script:
             }
           });
         }
+
+        function setSignatureWithInlineImage() {
```
emtencza, line 66: Should we mention in the snippet and in the body.setSignature snippet that this only work in win32?

ElizabethSamuel-MSFT: Actually, it worked when I tried in OWA. However, I'll add a comment to link folks to the Signature section of the preview requirement set page so users see how to try this feature.

ElizabethSamuel-MSFT: I opted to comment that setSignatureAsync is in preview because I'm mapping to released API (add attachment). I'm not doing so for the other samples because they're in the preview API snippets section and will only show in preview docs.

PR #473, opened 2020-08-05
[excel] (Range) Dynamic array sample

AlexJerabek:
Here's the gist link so you can try it out in Script Lab: https://gist.github.com/AlexJerabek/8889dee05873b05f85175e98ff38fd63

This sample shows how to use the new APIs around Range spilling to handle dynamic array content.

Note that this still needs the updates to the Excel mapping workbook for reference snippets. It is waiting on #472 to be checked in to avoid merge conflicts.


samples/excel/85-preview-apis/dynamic-arrays.yaml
```diff
@@ -0,0 +1,185 @@
...
+          } catch (error) {
+            // Note: In a production add-in, you'd want to notify the user through your add-in's UI.
+            console.error(error);
+          }
+        }
+    language: typescript
+template:
+    content: |-
+        <section class="ms-font-m">
+          <p>This sample shows how to apply formulas that return dynamic arrays and show how to get the relevant information
```
alison-mk, line 133:
```suggestion
          <p>This sample shows how to apply formulas that return dynamic arrays and how to get the relevant information
```

alison-mk: Intro sentence is unclear to me, is the second 'show' necessary? If so, what is doing the showing?

samples/excel/85-preview-apis/dynamic-arrays.yaml
```diff
@@ -0,0 +1,185 @@
...
+          <h3>Try it out</h3>
+          <button id="copy-table-headers" class="ms-Button">
+                  <span class="ms-Button-label">Copy table headers</span>
+              </button>
+          <p />
+          <button id="apply-filter-function" class="ms-Button">
+                <span class="ms-Button-label">Apply FILTER formula</span>
+            </button>
+          <p />
+          <p>Select a cell and press the following button to see any spill information for that cell related to dynamic
```
alison-mk, line 154: I think we should clarify that the user needs to select a new, spilled cell for this to work. Maybe 'Select a spilled cell'?

alison-mk, line 154:
```suggestion
          <p>Select a spilled cell and press the following button to see information for that cell related to dynamic
```

samples/excel/85-preview-apis/dynamic-arrays.yaml
```diff
@@ -0,0 +1,185 @@
...
+            arrays.</p>
+          <button id="display-spill-information" class="ms-Button">
+              <span class="ms-Button-label">Display spill information for selected range</span>
```
alison-mk, line 157: Since this button only works when a single cell is selected, should the button say 'selected range'? I'm still learning Excel terms so not sure! Along the same lines, should the console readouts say 'selected range' when we're only referring to a single cell?

samples/excel/85-preview-apis/dynamic-arrays.yaml
AlexJerabek: Nope! Good catch.

samples/excel/85-preview-apis/dynamic-arrays.yaml
AlexJerabek: I'll rephrase this to be more clear. It is a constant source of tension in the writing that a range can be a single cell.

PR #481, opened 2020-08-19
Adding Culture info and Direct precedents code samples

alison-mk:


samples/excel/85-preview-apis/culture-info-date-time.yaml
```diff
@@ -0,0 +1,138 @@
+order: 5
+id: excel-culture-info-date-time
+name: 'Culture info: date and time'
+description: This sample shows how to apply the cultural settings APIs to retrieve read-only operating system date and time settings.
```
AlexJerabek, line 4: It is still the operating system date/time when accessed through Excel for the Web? I think that's from the OneDrive/SharePoint site hosting the spreadsheet, but I could be wrong.

samples/excel/85-preview-apis/culture-info-date-time.yaml
```diff
@@ -0,0 +1,138 @@
...
+        $("#display-date-time-setting").click(() => tryCatch(displayDateTimeSetting));
+        $("#write-date-time-setting").click(() => tryCatch(writeDateTimeSetting));
+
+        /* Create table with only header content. */
+        async function setup() {
+          await Excel.run(async (context) => {
+            context.workbook.worksheets.getItemOrNullObject("Sample").delete();
+            const sheet = context.workbook.worksheets.add("Sample");
+            const range = sheet.getRange("A1:B6");
+            range.values = [["Culture Setting", "Setting Format"], ["", ""], ["", ""], ["", ""], ["", ""], ["", ""]];
```
AlexJerabek, line 20: Instead of adding several empty cells, considering making the original range "A1:B1" so you only need to worry about the headers.

samples/excel/85-preview-apis/culture-info-date-time.yaml
```diff
@@ -0,0 +1,138 @@
...
+api_set:
+    ExcelAPI: '1.12'
+script:
+    content: |
+        $("#setup").click(() => tryCatch(setup));
+        $("#display-date-time-setting").click(() => tryCatch(displayDateTimeSetting));
+        $("#write-date-time-setting").click(() => tryCatch(writeDateTimeSetting));
+
+        /* Create table with only header content. */
+        async function setup() {
```
AlexJerabek, line 15: Nit: The `setup` method is typically after the featured methods of the snippet (right above `tryCatch`).

samples/excel/85-preview-apis/culture-info-date-time.yaml
```diff
@@ -0,0 +1,138 @@
...
+            range.format.autofitColumns();
+            sheet.activate();
+            await context.sync();
+          });
+        }
+
+        /* Use the cultural settings API to display date and time settings in your console. */
+        async function displayDateTimeSetting() {
+          await Excel.run(async (context) => {
+            context.application.cultureInfo.datetimeFormat.load([
```
AlexJerabek, line 31: Consider adding a comment like "Load the date and time settings from the workbook."

snippet-extractor-output/snippets.yaml
```diff
@@ -1,7 +1,7 @@
 'Excel.Application#suspendScreenUpdatingUntilNextSync:member(1)':
   - >-
     // Link to full sample:
-    https://raw.githubusercontent.com/OfficeDev/office-js-snippets/master/samples/excel/90-scenarios/performance-optimization.yaml
+    https://raw.githubusercontent.com/OfficeDev/office-js-snippets/master/samples/office-js-snippets/samples/excel/90-scenarios/performance-optimization.yaml
```
AlexJerabek, line 4: Why did all the URLs in this file change? We should investigate that, because that doesn't seem right.

samples/excel/85-preview-apis/culture-info-date-time.yaml
```diff
@@ -0,0 +1,138 @@
...
+            const range = sheet.getRange("A1:B6");
+            range.values = [["Culture Setting", "Setting Format"], ["", ""], ["", ""], ["", ""], ["", ""], ["", ""]];
+            const table = sheet.tables.add("A1:B6", true);
+            range.format.autofitColumns();
+            sheet.activate();
+            await context.sync();
+          });
+        }
+
+        /* Use the cultural settings API to display date and time settings in your console. */
```
AlexJerabek, line 28: Consider pulling these comments into the `run` function. That way, they'll get brought along when the snippet extractor puts this code in the reference docs.

samples/excel/85-preview-apis/culture-info-date-time.yaml
```diff
@@ -0,0 +1,138 @@
...
+            const longDate = sheet.getRange("A2:B2");
+            longDate.values = [["Long date", systemLongDatePattern]];
+            const shortDate = sheet.getRange("A3:B3");
+            shortDate.values = [["Short date", systemShortDatePattern]];
+            const shortDate = sheet.getRange("A4:B4");
+            shortDate.values = [["Date separator", systemDateSeparator]];
+            const shortDate = sheet.getRange("A5:B5");
+            shortDate.values = [["Long time format", systemLongTimePattern]];
+            const shortDate = sheet.getRange("A6:B6");
+            shortDate.values = [["Time separator", systemTimeSeparator]];
```
AlexJerabek, line 87: While it doesn't matter at this scale, there is a best practice to reduce the number of Range objects created. Consider rewriting this to be one `getRange` operation and one set `values` operation.

samples/excel/85-preview-apis/culture-info-date-time.yaml
```diff
@@ -0,0 +1,138 @@
...
+            await callback();
+          } catch (error) {
+            // Note: In a production add-in, you'd want to notify the user through your add-in's UI.
+            console.error(error);
+          }
+        }
+    language: typescript
+template:
+    content: |-
+        -| <section class="ms-font-m"> <p>This sample shows how to apply the cultural settings APIs to retrieve read-only operating system date and time settings.</p> </section>
```
AlexJerabek, line 110: I think the "-|" at the beginning of this line is unintentional and can be removed.

samples/excel/85-preview-apis/culture-info-date-time.yaml
```diff
@@ -0,0 +1,138 @@
...
+        <section class="samples ms-font-m"> <h3>Setup</h3> <button id="setup" class="ms-Button"> <span class="ms-Button-label">Setup</span> </button> </section>
+        <section class="samples ms-font-m"> <h3>Try it out</h3> <button id="display-date-time-setting" class="ms-Button"> <span class="ms-Button-label">Display date/time settings</span> </button> <p></p> <button id="write-date-time-setting" class="ms-Button"> <span class="ms-Button-label">Write date/time settings</span> </button> </section>
```
AlexJerabek, line 112: Can we spread this out into more lines? You might want to copy-and-paste this section from Script Lab, since the export feature doesn't handle the HTML section well.

samples/excel/85-preview-apis/direct-precedents.yaml
```diff
@@ -0,0 +1,154 @@
+order: 6
+id: excel-direct-precedents
+name: Direct precedents
+description: 'This sample shows how to highlight precedent cells which directly affect the value of the currently selected cell, using the direct precedents API.'
```
AlexJerabek, line 4: Consider rephrasing this to explicitly mention formulas.
```suggestion
description: 'This sample shows how to find and highlight the precedents of the currently selected cell. Precedents  are cells referenced by the formula in a cell.'
```

samples/excel/85-preview-apis/direct-precedents.yaml
```diff
@@ -0,0 +1,154 @@
...
+          });
+        }
+
+        /* Select a cell with direct precedents on one worksheet. */
+        async function selectFirstCell() {
+          await Excel.run(async (context) => {
+            // Ensure correct worksheet is active.
+            const sheet = context.workbook.worksheets.getItem("Sample2020Data");
+            sheet.activate();
+            sheet.load("Sample2020Data");
```
AlexJerabek, line 70: This `load` statement is unnecessary.

samples/excel/85-preview-apis/direct-precedents.yaml
```diff
@@ -0,0 +1,154 @@
...
+          });
+        }
+
+        /* Select a cell with direct precedents on both worksheets. */
+        async function selectSecondCell() {
+          await Excel.run(async (context) => {
+            // Ensure correct worksheet is active.
+            const sheet = context.workbook.worksheets.getItem("Sample2020Data");
+            sheet.activate();
+            sheet.load("Sample2020Data");
```
AlexJerabek, line 85: This `load` statement is unnecessary.

samples/excel/85-preview-apis/direct-precedents.yaml
```diff
@@ -0,0 +1,154 @@
...
+
+        /* Select a cell with direct precedents on one worksheet. */
+        async function selectFirstCell() {
+          await Excel.run(async (context) => {
+            // Ensure correct worksheet is active.
+            const sheet = context.workbook.worksheets.getItem("Sample2020Data");
+            sheet.activate();
+            sheet.load("Sample2020Data");
+
+            // Select the cell.
```
AlexJerabek, line 72: Consider explicitly calling out which cells this cell references.

samples/excel/85-preview-apis/direct-precedents.yaml
```diff
@@ -0,0 +1,154 @@
...
+script:
+    content: |
+        $("#setup").click(() => tryCatch(setup));
+        $("#select-first-cell").click(() => tryCatch(selectFirstCell));
+        $("#select-second-cell").click(() => tryCatch(selectSecondCell));
+        $("#get-direct-precedents").click(() => tryCatch(getDirectPrecedents));
+        $("#clear-highlighting").click(() => tryCatch(setup));
+
+        /* Create two sample tables with functions that span the tables. */
+        async function setup() {
```
AlexJerabek, line 17: Same comment about `setup` placement.

samples/excel/85-preview-apis/direct-precedents.yaml
```diff
@@ -0,0 +1,154 @@
...
+            await callback();
+          } catch (error) {
+            // Note: In a production add-in, you'd want to notify the user through your add-in's UI.
+            console.error(error);
+          }
+        }
+    language: typescript
+template:
+    content: |-
+        -| <section class="ms-font-m"> <p>This sample shows how to highlight precedent cells which directly affect the value of the currently selected cell, using the direct precedents API.</p> </section>
```
AlexJerabek, line 126: Same comments as the culture info sample about this HTML section.

samples/excel/85-preview-apis/direct-precedents.yaml
```diff
@@ -0,0 +1,154 @@
...
+host: EXCEL
+api_set:
+    ExcelAPI: '1.12'
+script:
+    content: |
+        $("#setup").click(() => tryCatch(setup));
+        $("#select-first-cell").click(() => tryCatch(selectFirstCell));
+        $("#select-second-cell").click(() => tryCatch(selectSecondCell));
+        $("#get-direct-precedents").click(() => tryCatch(getDirectPrecedents));
+        $("#clear-highlighting").click(() => tryCatch(setup));
```
AlexJerabek, line 14: Let's write a separate method to clear the formatting. I think it'll be more clear to the people reading the sample, plus it might help teach someone how to clear formatting. You can get the used data ranges for each worksheet (`Worksheet.getUsedRange()`) and then clear the fill on both (Range.format.fill.clear()).

Happy to write this up in Script Lab for you, if you'd like.

samples/excel/85-preview-apis/direct-precedents.yaml
```diff
@@ -0,0 +1,154 @@
...
+            await context.sync();
+          });
+        }
+
+        /* Use the direct precedents API to detect precedent cells. */
+        async function getDirectPrecedents() {
+          await Excel.run(async (context) => {
+            let range = context.workbook.getActiveCell();
+            var rac = range.getDirectPrecedents();
+            range.load();
```
AlexJerabek, line 99: It's a best practice to only load what you need. Since the only property you're using is `address`, we can rewrite this to `range.load("address");`.

samples/excel/85-preview-apis/direct-precedents.yaml
```diff
@@ -0,0 +1,154 @@
...
+            rac.areas.load();
```
AlexJerabek, line 100: Same loading advice goes for this collection. You can say `rac.areas.load("address");` to load the address property for each member of the collection.

samples/excel/85-preview-apis/direct-precedents.yaml
```diff
@@ -0,0 +1,154 @@
...
+            range.select();
+            await context.sync();
+          });
+        }
+
+        /* Use the direct precedents API to detect precedent cells. */
+        async function getDirectPrecedents() {
+          await Excel.run(async (context) => {
+            let range = context.workbook.getActiveCell();
+            var rac = range.getDirectPrecedents();
```
AlexJerabek, line 98: Can we rename `rac` to something more informative? Maybe `directPrecedents`? Also, please change `var` to `let` as per TypeScript guidelines.

AlexJerabek, line 98: Let's add a concept here with a quick explanation of what a precedent is. That way, when this method gets used in the reference docs, it can work completely independently of the larger snippet.

samples/excel/85-preview-apis/culture-info-date-time.yaml
alison-mk: That way seems weird to me because the methods don't follow the same order in which they're triggered at the top of the file, but I think consistency across samples is important, so I will change this.

snippet-extractor-output/snippets.yaml
alison-mk: I have no idea but I can just undo this and keep an eye out for it in the future.

samples/excel/85-preview-apis/culture-info-date-time.yaml
alison-mk: Good question! How should I look into this? The feature spec says "This is based on current system settings" in the description for each of the new dateTimeFormat APIs. Would it be better if I simply dropped 'operating' from the sample description + other references in this sample?

samples/excel/85-preview-apis/culture-info-date-time.yaml
alison-mk: Since these are system settings (based on all the info I have), would you agree with "Load the system date and time settings." as a comment instead?

samples/excel/85-preview-apis/culture-info-date-time.yaml
alison-mk: Something happened with the formatting -- I explicitly followed your directions about copy/pasting from the HTML panel, but somehow this is all still minified, even in my local version. I wonder if it happened when I ran yarn start? Or maybe something with my VSCode... any ideas? It also added the extra -| on both files.

samples/excel/85-preview-apis/culture-info-date-time.yaml
AlexJerabek: Yeah, that seems like a reasonable solution.

samples/excel/85-preview-apis/culture-info-date-time.yaml
AlexJerabek: Yeah, that works!

samples/excel/85-preview-apis/direct-precedents.yaml
alison-mk: Ok, thanks! Can you explain why?

samples/excel/85-preview-apis/culture-info-date-time.yaml
AlexJerabek: It might be something happening during yarn start. We can take a deeper look or fix it in a subsequent PR.

samples/excel/85-preview-apis/direct-precedents.yaml
alison-mk: Ok, would love to see how you write this. I considered writing a new method to clear the formatting at the time, but it seemed unnecessary. I agree with your concerns/suggestions though.

samples/excel/85-preview-apis/direct-precedents.yaml
alison-mk: Sure! This block is directly from the spec code sample. Can you point me towards a good TypeScript guidelines reference?

samples/excel/85-preview-apis/direct-precedents.yaml
AlexJerabek: Sure!

You don't need a load because you're not actually getting any data from the workbook. The `getRange` method is just creating a proxy object that reflects that range. It's a reference to a range, not the contents of that range. To get values from it, you'd need to call `myRange.load("values")`. The same logic applies to worksheets. You're getting a reference to a worksheet named "Sample2020Data", not the worksheet itself. The snippet doesn't check if that worksheet exists until the `context.sync` call. Effectively, your `selectFirstCell` method is a series of reference creations and write operations (`activate` and `select`). It never calls anything that'll read from the workbook (from the add-in's perspective).

Relatedly, `load` queues up a command to pull property values from the workbook to your add-in. The string argument is a property name. In this case, by passing something that's not a property name, it's actually not loading anything.

samples/excel/85-preview-apis/direct-precedents.yaml
AlexJerabek: Here's a thorough description of `var`, `let`, and `const`: https://www.typescriptlang.org/docs/handbook/variable-declarations.html

Basically, there's some small scoping problems a user can encounter with `var` that are avoided by using `let`. Since we don't know how people are going to copy-paste our code, we want to try and make the most stable blocks we can.

samples/excel/85-preview-apis/culture-info-date-time.yaml
```diff
@@ -0,0 +1,138 @@
...
+            context.application.cultureInfo.datetimeFormat.load([
+              "longDatePattern",
+              "shortDatePattern",
+              "dateSeparator",
+              "longTimePattern",
+              "timeSeparator"
+            ]);
+            await context.sync();
+
+            // System settings are read-only and set in the user's operating system.
```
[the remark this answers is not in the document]
alison-mk: Thank you for pointing this out! The settings are read-only in Excel, like the add-in developer can only access them, not change them. As I understand it, a computer owner/user is the only person who can change their system date and time settings. Based on this and previous feedback from Alex, I will reformat the comments and adjust the phrasing here to address your concern. If you could take a look again when I push updates, I would appreciate it!

samples/excel/85-preview-apis/culture-info-date-time.yaml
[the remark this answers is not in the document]
alison-mk: Could you explain the logic here? Is it the APIs that are read-only, not the settings themselves? I really don't know.

PR #482, opened 2020-08-21
[excel] (PivotTable) Add PivotFilter sample

AlexJerabek:
This sample demonstrates the new PivotFilter functionality. It walks users through the different types of filter with a common example for each.

Note that this PR does not yet include the snippet metadata for the reference docs. I'll wait for #481 to get checked to avoid conflicts.


AlexJerabek:
@alison-mk The stability of this sample appears to have improved. Can you please verify before I check in?

samples/excel/85-preview-apis/pivottable-pivotfilters.yaml
```diff
@@ -0,0 +1,326 @@
...
+          <button id="manual-filter" class="ms-Button">
+              <span class="ms-Button-label">Organic filter (Manual)</span>
+          </button>
+          <p />
+          <button id="date-filter" class="ms-Button">
+              <span class="ms-Button-label">Before August filter (Date)</span>
+          </button>
+          <p />
+          <button id="value-filter" class="ms-Button">
+              <span class="ms-Button-label">>500 units wholesale filter (Value)</span>
```
alison-mk, line 284: The button labels are a little confusing to me. I think it's because two of them are exclusive (removing events before August and removing lemons + limes) while the other is inclusive (only showing rows with more than 500 wholesale crates). At first I thought the sample wasn't working correctly, but after investigation I think tweaking the button labels would solve my confusion. Maybe something like: "Filter out events before August", "Filter out rows with less than 500 units wholesale", and "Filter out lemons/limes"

alison-mk (approved):
No more Excel crashes - this sample works for me now! 👏 🥳 
One comment below about clarifying the text for some buttons.

AlexJerabek:
Thanks @alison-mk! You're right that the filters are inconsistent in how they filter. I did a pass to try and be more consistent and clear. How does this look?

samples/excel/85-preview-apis/pivottable-pivotfilters.yaml
```diff
@@ -55,10 +55,10 @@ script:
               dateHierarchy = pivotTable.rowHierarchies.add(pivotTable.hierarchies.getItem("Date Updated"));
             }
 
-            // Apply a date filter for anything logged before August.
+            // Apply a date filter to fliter out anything logged before August.
```
alison-mk, line 58:
```suggestion
            // Apply a date filter to filter out anything logged before August.
```

samples/excel/85-preview-apis/pivottable-pivotfilters.yaml
```diff
@@ -55,10 +55,10 @@ script:
...
             const filterField = dateHierarchy.fields.getItem("Date Updated");
             const dateFilter = {
-              condition: Excel.DateFilterCondition.before,
+              condition: Excel.DateFilterCondition.afterOrEqualTo,
```
alison-mk, line 61: I think something is off here, clicking the August 2020 button is also displaying a bunch of July dates.

alison-mk (commented):
Buttons + additional descriptive text looks great! I found another bug though, noted below ☹️

AlexJerabek:
The `DateFilter` bug that @alison-mk is reporting should be fixed by Build 13304.

PR #484, opened 2020-08-26
Fix two minor errors in direct precedents sample

alison-mk:


alison-mk:
Ignore this PR! Will open a new PR momentarily.

PR #488, opened 2020-09-18
[excel] (Events) add ExcelApi 1.12 comment event handler sample

alison-mk:


samples/excel/30-events/events-comment-event-handler.yaml
```diff
@@ -0,0 +1,195 @@
+order: 13
+id: excel-events-comment-event-handler
```
AlexJerabek, line 2: This is outside of the scope of this PR, but we should reconsider if an "events" category makes sense. It might be a better organizational tactic to put the comment events snippet with the other comment samples (same for charts, tables, and the rest). Again, that's beyond this PR, but something to think about.

samples/excel/30-events/events-comment-event-handler.yaml
```diff
@@ -0,0 +1,195 @@
...
+        <section class="ms-font-m">
+            <p>This sample shows how to add, edit, and remove comments.</p>
+        </section>
+        <section class="setup ms-font-m">
+            <h3>Setup</h3>
+            <button id="setup" class="ms-Button">
+                <span class="ms-Button-label">Create a worksheet</span>
+            </button>
+            <p></p>
+            <button id="registerEventHandlers" class="ms-Button">
```
AlexJerabek, line 146: This ID doesn't follow the kebab casing your other IDs do (`register-event-handlers`).
(Note that if you change this here, you'll need to change the click listener in the content section.)

samples/excel/30-events/events-comment-event-handler.yaml
```diff
@@ -0,0 +1,195 @@
...
+                const comments = context.workbook.worksheets.getActiveWorksheet().comments;
+
+                // Register the onAdded, onChanged, and onDeleted comment event handlers.
+                comments.onAdded.add(commentAdded);
+                comments.onChanged.add(commentChanged);
+                comments.onDeleted.add(commentDeleted);
+
+                await context.sync();
+
+                console.log("Added handlers for Comment onAdded, onChanged, and onDeleted events.");
```
AlexJerabek, line 28: Technically, this is for the CommentCollection.on* events. I think we should either reference the proper object structure or rephrase this to be clear you're talking about the behavior in conceptual terms, rather than programmatic ones.

samples/excel/30-events/events-comment-event-handler.yaml
```diff
@@ -0,0 +1,195 @@
...
+            });
+        }
+
+        async function commentAdded(event: Excel.CommentAddedEventArgs) {
+            await Excel.run(async (context) => {
+                // Retrieve the added comment.
+                let addedComment = context.workbook.comments.getItem(event.commentDetails[0].commentId);
```
AlexJerabek, line 35: This snippet assumes only a single comment is getting added/deleted/changed at a time. Consider either expanding the sample to work work (and showcase) multi-comment events or add comments explaining the limitation.
(This note applies to all three listeners.)

samples/excel/30-events/events-comment-event-handler.yaml
```diff
@@ -0,0 +1,195 @@
...
+
+                // Load the added comment's data.
+                addedComment.load(["content", "authorName", "creationDate"]);
+
+                await context.sync();
+
+                // Print out the added comment's data.
+                console.log(`A comment was added:`);
+                console.log(`    ID: ${event.commentDetails[0].commentId}`);
```
AlexJerabek, line 44: Instead of using the actual tab character, use `\t`. That will display consistently across browsers and IDEs. It also makes it so you can tell the difference between spaces and tabs at a quick glance.

samples/excel/30-events/events-comment-event-handler.yaml
```diff
@@ -0,0 +1,195 @@
...
+
+                await context.sync();
+
+                console.log("Added handlers for Comment onAdded, onChanged, and onDeleted events.");
+            });
+        }
+
+        async function commentAdded(event: Excel.CommentAddedEventArgs) {
+            await Excel.run(async (context) => {
+                // Retrieve the added comment.
```
AlexJerabek, line 34: Consider pointing out that the ID is what is being used to map the event to the comment.
```suggestion
                // Retrieve the added comment using the comment ID from the event args.
```

(Note that if you make this or a related change, you should adjust the comment in the other event listeners.)

samples/excel/30-events/events-comment-event-handler.yaml
AlexJerabek, line 35: Nit: You can use `const` instead of `let` since you're not changing the value of the comment.

samples/excel/30-events/events-comment-event-handler.yaml
```diff
@@ -0,0 +1,195 @@
...
+                const selectedRange = context.workbook.getSelectedRange();
+
+                // Note that an InvalidArgument error will be thrown if multiple cells are selected.
+                context.workbook.comments.add(selectedRange, "TODO: add headers here.");
+
+                await context.sync();
+            });
+        }
+
+        async function addCommentToCell() {
```
AlexJerabek, line 89: The original comment sample had two "add comment" methods to showcase using either a range or an address as the entry point of the comment. I'm not sure having two add functions adds value to this sample. You can get by fine with just one.

samples/excel/30-events/events-comment-event-handler.yaml
```diff
@@ -0,0 +1,195 @@
...
+            await Excel.run(async (context) => {
+                const sheet = context.workbook.worksheets.getItem("Comments");
+
+                // Note that an InvalidArgument error will be thrown if multiple cells are passed to `comment.add`.
+                sheet.comments.add("A2", "TODO: add data.");
+                await context.sync();
+            });
+        }
+
+        async function editComment() {
```
AlexJerabek, line 99: Consider as a future work item adding comment replies to this sample. I think it would be interesting to show how adding, deleting, and editing replies triggers these events.

samples/excel/30-events/events-comment-event-handler.yaml
```diff
@@ -0,0 +1,195 @@
...
+            } catch (error) {
+                // Note: In a production add-in, you'd want to notify the user through your add-in's UI.
+                console.error(error);
+            }
+        }
+    language: typescript
+template:
+    content: |-
+        <section class="ms-font-m">
+            <p>This sample shows how to add, edit, and remove comments.</p>
```
AlexJerabek, line 138: Please update this comment to reflect the sample's goal of showing comment-related events.

samples/excel/30-events/events-comment-event-handler.yaml
```diff
@@ -0,0 +1,195 @@
...
+                <span class="ms-Button-label">Create a worksheet</span>
+            </button>
+            <p></p>
+            <button id="registerEventHandlers" class="ms-Button">
+                <span class="ms-Button-label">Register event handlers</span>
+            </button>
+        </section>
+        <section class="samples ms-font-m">
+            <h3>Try it out</h3>
+            <p>To better view the comment changes, open the Comments pane by selecting <b>Show Comments</b> from the
```
AlexJerabek, line 152: I'm not sure this note is as interesting in this sample, versus the original comment basics one. Your call on whether to keep it.

samples/excel/30-events/events-comment-event-handler.yaml
```diff
@@ -0,0 +1,195 @@
...
+            <button id="add-comment-to-selected-cell" class="ms-Button">
+                <span class="ms-Button-label">Add comment to selected cell</span>
+            </button>
+            <p></p>
+            <button id="add-comment-to-cell" class="ms-Button">
+                <span class="ms-Button-label">Add comment to A2</span>
+            </button>
+            <p></p>
+            <button id="edit-comment" class="ms-Button">
+                <span class="ms-Button-label">Edit first comment</span>
```
AlexJerabek, line 163: If you take the earlier note of dialing down to one "add" action, we can simplify these buttons to just be "Add comment", "Edit comment", and "Remove comment", instead of giving more specifics.

samples/excel/30-events/events-comment-event-handler.yaml
AlexJerabek, line 2: Consider changing the ID to better match the other event sample IDs.
```suggestion
id: excel-events-comments
```
(Note that if you make this change, you'll need to edit the excel mapping file too.)

AlexJerabek (changes requested):
Looks good @alison-mk. I left some suggestions for streamlining the sample and other little things.

samples/excel/30-events/events-comment-event-handler.yaml
alison-mk: I had the same thought while selecting the appropriate folder for this new sample. I put the sample in this folder because it's the most logical location while the 'events' category exists, but don't think an 'events' category is particularly helpful, and I do want to reorganize the samples in this folder in the future. Thanks for mentioning this!

samples/excel/30-events/events-comment-event-handler.yaml
alison-mk: My main concern about this sentence is that the Comments pane conflicts with the Script Lab task pane -- a user needs to click between two overlapping panes repeatedly to see the comment changes (in either sample). I left the sentence in here because it might be helpful for a developer less familiar with the little purple comments flag that appears on cells with comments.

samples/excel/30-events/events-comment-event-handler.yaml
alison-mk: Per the yarn errors, I switched this file to use spaces instead of tabs. The console lines are all using an indentation of 4 spaces (I prefer 4-space to 2-space indentations, for readability). Using `\t` makes the script code less readable in my opinion, so I would rather use spaces. 

I have seen other samples use spaces for indentation similar to this implementation, so one of my goals was to be consistent with those. I haven't seen `\t` in any samples yet (but I also haven't seen all the samples). 

Are there issues with using spaces across browsers and IDEs?

AlexJerabek: Using spaces is fine.

samples/excel/30-events/events-comment-event-handler.yaml
alison-mk: I implemented part of this. Thanks for the suggestion!

samples/excel/30-events/events-comment-event-handler.yaml
alison-mk: I adjusted this sentence to better clarify its purpose in this sample (to make sure the user is aware of the Comments pane).

samples/excel/30-events/events-comment-event-handler.yaml
alison-mk: I added comments to point this out, let me know if you think they're sufficient.

AlexJerabek (approved):

alison-mk:
Thank you for all the suggestions @AlexJerabek!

PR #491, opened 2020-09-19
[excel] (Charts) Bubble chart sample

AlexJerabek:
There wasn't a good bubble chart sample, so I made one. This shows how to fine tune the chart series to get an aesthetically pleasing bubble chart that conveys four columns of data on a 2D plane.


alison-mk (approved):
Looks great! Very fun fruit selection. Two minor comments below.

AlexJerabek:
Thanks for the review @alison-mk! And thank you to the humble citrus genus, for making the world a brighter place.

PR #492, opened 2020-09-21
[admin] (Excel snippet mapping) Removing extra mapping line

AlexJerabek:
There was a line for `Range.workbookRangeAreas` that didn't map to an API. I'm assuming it was accidentally added and so I'm removing it. If this is supposed to go to a different API, please let me know.


alison-mk (approved):
Thank you for catching this!

PR #495, opened 2020-09-25
[Outlook] (delegate) Update snippet with appointment example

ElizabethSamuel-MSFT:


samples/outlook/65-delegates-and-shared-folders/get-shared-properties.yaml
```diff
@@ -61,6 +62,48 @@ script:
...
+              Office.context.mailbox.item.getSharedPropertiesAsync(
+                {
+                  // Pass auth token along.
+                  asyncContext: result.value
+                },
+                function(result2) {
+                  let sharedProperties = result2.value;
+                  let delegatePermissions = sharedProperties.delegatePermissions;
+
+                  // Determine if user can do the expected operation, e.g., do they have Write permission?
```
Oleg-O, line 83: I guess this is existing on the one you're copying from... but this comment is a bit confusing - we say "Write" but really check for Read permissions in the code below. Update the comment to "Read" too?

Oleg-O (approved):

samples/outlook/65-delegates-and-shared-folders/get-shared-properties.yaml
ElizabethSamuel-MSFT: Updated. Thx.

PR #496, opened 2020-09-29
[excel] (PivotTable) Removing check for isFiltered

AlexJerabek:
The calls to `isFitlered` are both unnecessary and incorrect. That method returns a `ClientResult`, which would have required a `context.sync` to read the value of (but the sample wasn't reading the value, just the `ClientResult` object itself).


alison-mk (approved):
Looks good! 
Note: The date filter still doesn't work properly on my build of Excel.

AlexJerabek:
@alison-mk I'll forward you an internal thread on the subject that might be related.

PR #526, opened 2021-03-06
[Excel] Update the add file from base64 snippet

alison-mk:
Aligning existing snippet with new Preview API.


samples/excel/85-preview-apis/workbook-insert-external-worksheets.yaml
```diff
@@ -15,21 +15,26 @@ script:
...
-                        );
+                    
+                    // Retrieve the workbook.
+                    const workbook = context.workbook;
+                    
+                    // Set up the insert options. 
+                    var options = { 
+                        sheetNamesToInsert: [], // Insert all the worksheets from the source workbook.
+                        positionType: Excel.WorksheetPositionType.after, // Insert after the `relativeTo` sheet.
+                        relativeTo: "Sheet1" }; // A worksheet name. Use with `positionType` to specify insert location.
```
AlexJerabek, line 29: The other argument comments seem specific to the sample scenario, so consider framing this one like that too.
```suggestion
                        relativeTo: "Sheet1" }; // The sheet relative to which the other worksheets will be inserted.
```

AlexJerabek (approved):
Approving with one suggestion.

PR #536, opened 2021-05-11
[Excel] Add direct dependents preview sample

alison-mk:
Adding ExcelApi 1.13 getDirectDependents code sample.


samples/excel/85-preview-apis/range-direct-dependents.yaml
```diff
@@ -0,0 +1,186 @@
...
+    $("#setup").click(() => tryCatch(setup));
+    $("#select-first-cell").click(() => tryCatch(selectFirstCell));
+    $("#select-second-cell").click(() => tryCatch(selectSecondCell));
+    $("#get-direct-dependents").click(() => tryCatch(getDirectDependents));
+    $("#clear-highlighting").click(() => tryCatch(clearFormatting));
+    
+    /** Select a cell with direct dependents on the same worksheet. */
+    async function selectFirstCell() {
+      await Excel.run(async (context) => {
+        // Ensure correct worksheet is active.
```
AlexJerabek, line 19: Consider making this more direct (if you take this suggestion, please apply it to the other selection method).
```suggestion
        // Activate the sample worksheet.
```

samples/excel/85-preview-apis/range-direct-dependents.yaml
```diff
@@ -0,0 +1,186 @@
+order: 4
+id: excel-direct-dependents
```
AlexJerabek, line 2: Since these APIs are in preview, a sample separate from the precedents API makes sense. We should consider merging them down the line, since the APIs are closely related and should probably be discovered at the same time.

AlexJerabek (approved):
Looks good. Approving with suggestions.

samples/excel/85-preview-apis/range-direct-dependents.yaml
alison-mk: I agree.

PR #537, opened 2021-05-11
[Excel] Add workbook activated preview sample

alison-mk:


samples/excel/85-preview-apis/events-workbook-activated.yaml
```diff
@@ -0,0 +1,80 @@
...
+        workbook.onActivated.add(workbookActivated);
+
+        await context.sync();
+        console.log("Added event handler for workbook activated.");
+      });
+    }
+
+    async function workbookActivated(event: Excel.WorkbookActivatedEventArgs) {
+      await Excel.run(async (context) => {
+        // Script that runs when the workbook is activated.
```
AlexJerabek, line 26: Let's avoid the term "script", since it's overloaded as is. Also, consider pulling this comment to above the `workbookActivated` function.
```suggestion
        // Callback function for when the workbook is activated.
```

samples/excel/85-preview-apis/events-workbook-activated.yaml
```diff
@@ -0,0 +1,80 @@
...
+        console.error(error);
+      }
+    }
+  language: typescript
+template:
+    content: |-
+      <section class="ms-font-m">
+        <p>This sample shows how to register a workbook activated event handler.</p>
+        <p>Once the event handler is registered, a notification will print to the console when the workbook is activated. Try
+          switching to another application and then switching back to Excel to see the console notification.</p>
```
AlexJerabek, line 46: Nit: Avoid "will" for brevity.
```suggestion
        <p>Once the event handler is registered, a notification prints to the console when the workbook is activated. Try
          switching to another application and then switching back to Excel to see the console notification.</p>
```

PR #540, opened 2021-05-12
[Excel] Add table resize preview sample

alison-mk:
Adding ExcelApi 1.13 table resize code sample.


AlexJerabek (approved):
Looks good. Two things to note:
- Please run `yarn start` before merging this change. That generates the snippet files needed for the reference docs.
- This might be a good candidate to merge with another snippet in the future.

PR #547, opened 2021-05-24
[Admin] main -> prod

alison-mk:


snippet-extractor-output/snippets.yaml
```diff
@@ -3065,6 +3065,52 @@
...
+      // This method acts like the Ctrl+Shift+Arrow key keyboard shortcut while a range is selected.
+      const extendedRange = range.getExtendedRange(
+        "Right", // Specify the direction as a string.
+        activeCell // If the selected range contains more than one cell, the active cell must be defined.
+      );
+      extendedRange.select();
+
+      await context.sync();
+    });
+  - >-
```
AlexJerabek, line 3092: Are you sure you want both these samples mapping to the same block in the ref docs? I think just one of them for getExtendedRange should be sufficient.

snippet-extractor-output/snippets.yaml
```diff
@@ -3179,6 +3225,54 @@
...
+        "Left", // Specify the direction as a string.
+        activeCell // If the selected range contains more than one cell, the active cell must be defined.
+      );
+
+      // Select the edge of the range.
+      rangeEdge.select();
+
+      await context.sync();
+    });
+  - >-
```
AlexJerabek, line 3252: Same comment as above for getRangeEdge.

AlexJerabek (approved):
Approving, but consider limiting the snippets to one per method.

snippet-extractor-output/snippets.yaml
alison-mk: I included both because they show different ways to specify the direction.

AlexJerabek: Let's see how it looks in the ref docs tomorrow. I don't think either way is bad.

PR #550, opened 2021-05-27
[Excel] Update insert worksheets sample to throw errors correctly

alison-mk:
This PR adds try-catch logic to the insert worksheets sample, ensuring that the proper error is thrown in Excel on the web when unsupported features are found in the external workbook. The previous iteration did not throw this error correctly. 
Note: Will update excel.xlsx metadata and run yarn start after incorporating feedback.


AlexJerabek (approved):
Approving, with some offline feedback about the global variable name.

PR #556, opened 2021-07-02
[Excel] Promote ExcelApi 1.13 snippets from preview to GA

alison-mk:


samples/excel/30-events/events-formula-changed.yaml
```diff
@@ -1,4 +1,4 @@
-order: 3
+order: 14
 id: excel-events-formula-changed
 name: Formula change event
```
ElizabethSamuel-MSFT, line 3:
```suggestion
name: Formula changed event
```

ElizabethSamuel-MSFT (approved):
If you accept this suggestion, remember to re-run yarn start before merging

samples/excel/30-events/events-formula-changed.yaml
alison-mk: I chose "Formula change event" as the name because it matches other snippet names that have change in the name: "Data change event", "Data change event details", and "Selection change events", but I agree that "changed" is more accurate (particularly because it matches the API names which all include "changed"). 

In conclusion, I'll accept your suggestion and adjust the names of the other three snippets to match. Thanks!

PR #578, opened 2021-11-24
[excel] (Chart) Remap data table snippet

AlexJerabek:
Minor thing I noticed while generatingthe reference docs: This snippet should point to `Chart.getDataTableOrNullObject`, instead of `ChartDataTable.getDataTableOrNullObject`.


alison-mk (approved):
Thank you!

PR #582, opened 2021-12-15
[Word] (preview) Footnotes and Comments features

ElizabethSamuel-MSFT:


samples/word/99-preview-apis/manage-comments.yaml
```diff
@@ -0,0 +1,235 @@
+id: word-manage-comments
+name: Manage comments
+description: Basic operations using comments.
```
AlexJerabek, line 3: Can we elaborate on this a bit? Any keywords here drive search, so maybe something enumerating the operations?

samples/word/99-preview-apis/manage-comments.yaml
```diff
@@ -0,0 +1,235 @@
...
+        $("#resolve").click(() => tryCatch(resolveFirstCommentInSelection));
+        $("#delete").click(() => tryCatch(deleteFirstCommentInSelection));
+        $("#range").click(() => tryCatch(getFirstCommentRange));
+        $("#setup").click(() => tryCatch(setup));
+
+        async function insertComment() {
+          await Word.run(async (context) => {
+            let text = $("#comment-text")
+              .val()
+              .toString();
```
AlexJerabek, line 23: Nit: Here and elsewhere, can this go on one line? IMO, it's harder to read like this.

samples/word/99-preview-apis/manage-comments.yaml
```diff
@@ -0,0 +1,235 @@
...
+            let comment = context.document.getSelection().insertComment(text);
+            comment.load("*");
```
AlexJerabek, line 25: We should avoid using the load everything command in samples ([source](https://docs.microsoft.com/en-us/office/dev/add-ins/develop/application-specific-api-model#load)). Let's only load specific properties for the `console.log` statement.

samples/word/99-preview-apis/manage-comments.yaml
```diff
@@ -0,0 +1,235 @@
...
+            await context.sync();
+            console.log("Comment inserted:");
+            console.log(comment);
```
AlexJerabek, line 28: Nit: Let's make this one `log` statement.

samples/word/99-preview-apis/manage-comments.yaml
```diff
@@ -0,0 +1,235 @@
...
+        $("#reply").click(() => tryCatch(replyToFirstCommentInSelection));
+        $("#get-comments").click(() => tryCatch(getComments));
+        $("#get-comments-in-selection").click(() => tryCatch(getCommentsInSelection));
+        $("#edit").click(() => tryCatch(editFirstCommentInSelection));
+        $("#resolve").click(() => tryCatch(resolveFirstCommentInSelection));
+        $("#delete").click(() => tryCatch(deleteFirstCommentInSelection));
+        $("#range").click(() => tryCatch(getFirstCommentRange));
+        $("#setup").click(() => tryCatch(setup));
+
+        async function insertComment() {
```
AlexJerabek, line 19: As per our recent code comment discussion, let's add comments before each method. Here's an Excel example to help with consistency: https://github.com/OfficeDev/office-js-snippets/blob/main/samples/excel/42-range/precedents.yaml

samples/word/99-preview-apis/manage-comments.yaml
```diff
@@ -0,0 +1,235 @@
...
+              .toString();
+            let comments = context.document.getSelection().getComments();
+            comments.load("*");
+            comments.load("items");
+            await context.sync();
+            let reply = comments.items[0].reply(text);
+            reply.load("*");
+            await context.sync();
+            console.log("Reply added:");
+            console.log(reply);
```
AlexJerabek, line 45: This seems like too many loads and syncs. I don't think you have to load the collection or the items.

samples/word/99-preview-apis/manage-comments.yaml
```diff
@@ -0,0 +1,235 @@
...
+        async function resolveFirstCommentInSelection() {
+          await Word.run(async (context) => {
+            let comment = context.document
+              .getSelection()
+              .getComments()
+              .getFirst();
+            comment.load("resolved");
+            await context.sync();
+
+            let test: Boolean = comment.resolved;
```
AlexJerabek, line 96: Why is `Boolean` capitalized? Shouldn't it be the primitive type `boolean`?

AlexJerabek, line 96: Also, this variable probably shouldn't be named `test`.

samples/word/99-preview-apis/manage-comments.yaml
```diff
@@ -0,0 +1,235 @@
+id: word-manage-comments
+name: Manage comments
```
AlexJerabek, line 2: Let's just name this "Comments".

samples/word/99-preview-apis/manage-comments.yaml
ElizabethSamuel-MSFT: Will do

samples/word/99-preview-apis/manage-comments.yaml
ElizabethSamuel-MSFT: This is how it was autoformatted. I had all of these on the same line.

samples/word/99-preview-apis/manage-footnotes.yaml
```diff
@@ -0,0 +1,264 @@
+id: word-manage-footnotes
+name: Manage footnotes
+description: Basic operations using footnotes.
```
AlexJerabek, line 3: Same name and description suggestions are the previous one: name it "Footnotes" and elaborate a bit more in the description.

samples/word/99-preview-apis/manage-comments.yaml
ElizabethSamuel-MSFT: Will update

samples/word/99-preview-apis/manage-footnotes.yaml
```diff
@@ -0,0 +1,264 @@
...
+          });
+        }
+        async function getFootnoteType() {
+          await Word.run(async (context) => {
+            const footnotes = context.document.body.footnotes;
+            context.load(footnotes);
+            await context.sync();
+
+            const reference_number = $("#input-reference").val();
+            let mark = Number(reference_number) - 1;
```
AlexJerabek, line 68: Why is a `Number` object being created with a constructor here?

samples/word/99-preview-apis/manage-footnotes.yaml
```diff
@@ -0,0 +1,264 @@
...
+        async function getReference() {
+          await Word.run(async (context) => {
+            const footnotes = context.document.body.footnotes;
+            context.load(footnotes);
+            await context.sync();
+
+            const reference_number = $("#input-reference").val();
+            let mark = Number(reference_number) - 1;
+            const item = footnotes.items[mark];
+            var reference = item.reference;
```
AlexJerabek, line 52: Try to avoid `var` in TS samples. Favor `let` or `const`.

AlexJerabek (changes requested):
Thanks @ElizabethSamuel-MSFT. I think these do a good job of showing the scenarios. I have concerns about the number of load/syncs and the times everything is loaded. I think this could be optimized to help teach the right, scalable patterns.

samples/word/99-preview-apis/manage-comments.yaml
ElizabethSamuel-MSFT: Ended up keeping the extra log statement because I preferred how the log is formatted in the Script Lab and its interactivity

samples/word/99-preview-apis/manage-comments.yaml
ElizabethSamuel-MSFT: Updated

samples/word/99-preview-apis/manage-footnotes.yaml
```diff
@@ -0,0 +1,266 @@
...
+
+            const reference_number = $("#input-reference").val();
+            let mark = (reference_number as number) - 1;
+            const item = footnotes.items[mark];
+            let reference = item.reference;
+            context.load(reference);
+            await context.sync();
+
+            reference.select();
+            await context.sync();
```
AlexJerabek, line 57: This can be written to only load the reference, instead of everything with a footnote. You can load specific properties of items in a collection like this:
```suggestion
      const footnotes = context.document.body.footnotes;
      footnotes.load("items/reference");
      await context.sync();

      const reference_number = $("#input-reference").val();
      let mark = (reference_number as number) - 1;
      const item = footnotes.items[mark];
      let reference = item.reference;
      reference.select();
      await context.sync();
```

samples/word/99-preview-apis/manage-footnotes.yaml
```diff
@@ -0,0 +1,266 @@
...
+
+            console.log(`Reference ${reference_number} is selected.`);
+          });
+        }
+        /** Get the referenced note's type, which is footnote. */
+        async function getFootnoteType() {
+          await Word.run(async (context) => {
+            const footnotes = context.document.body.footnotes;
+            context.load(footnotes);
```
AlexJerabek, line 66: Same load comment as earlier.

samples/word/99-preview-apis/manage-footnotes.yaml
```diff
@@ -0,0 +1,266 @@
...
+            await context.sync();
+
+            const reference_number = $("#input-reference").val();
+            let mark = (reference_number as number) - 1;
+            const item = footnotes.items[mark].getNext();
+            context.load(item);
+            await context.sync();
```
AlexJerabek, line 126: This load/sync isn't needed.

samples/word/99-preview-apis/manage-footnotes.yaml
```diff
@@ -0,0 +1,266 @@
...
+            const footnotes = context.document.body.footnotes;
+            context.load(footnotes);
+            await context.sync();
+
+            const reference_number = $("#input-reference").val();
+            let mark = (reference_number as number) - 1;
+            const item = footnotes.items[mark];
+
+            context.load(item);
+            await context.sync();
```
AlexJerabek, line 149: This load/sync isn't needed.

AlexJerabek (changes requested):
Thanks for making those changes Elizabeth. I think this reads a lot better. There are still some unnecessary load statements we should remove before publishing.

Also, I'm not sure if this is a big deal, but these samples use `context.load(myObject)` instead of `myObject.load(property)`. The latter is used in other samples and this article: https://docs.microsoft.com/en-us/office/dev/add-ins/develop/application-specific-api-model?view=word-js-preview#load
Might be worth having a chat about consistency in an upcoming team meeting, so we're aligned on the proper pattern for Word, PowerPoint, and Excel.

ElizabethSamuel-MSFT:
@AlexJerabek I'll go ahead and merge this. Do leave any further feedback and I'll address in a subsequent PR. Thx!

PR #584, opened 2022-01-13
[Excel] (Data types) Add preview sample for formatted number data types

alison-mk:
[Internal work item](https://office.visualstudio.com/OC/_workitems/edit/5458514).


samples/excel/85-preview-apis/data-types-formatted-number.yaml
```diff
@@ -0,0 +1,149 @@
...
+
+        async function setFormattedNumberDate() {
+          await Excel.run(async (context) => {
+            // This method creates a formatted number data type,
+            // and sets the format of this data type as a date.
+
+            // Create a worksheet called "Sample" and activate it.
+            context.workbook.worksheets.getItemOrNullObject("Sample").delete();
+            const sheet = context.workbook.worksheets.add("Sample");
+            sheet.activate();
```
AlexJerabek, line 22: When I was playing around with the sample, I missed this worksheet reset was here. Could this be moved to a "Setup" method and button? Alternatively, you could have the buttons target the currently selected cell, if you like the "setup-less" sample style.

samples/excel/85-preview-apis/data-types-formatted-number.yaml
```diff
@@ -0,0 +1,149 @@
+order: 1
+id: excel-data-types-formatted-number
+name: 'Data types: Formatted numbers'
+description: This sample shows how to set and get data types using the formatted number properties.
```
AlexJerabek, line 4: Meta question for the feature: why would I opt to use the formatted number data type instead of setting `values` and `numberFormats` separately?

samples/excel/85-preview-apis/data-types-formatted-number.yaml
```diff
@@ -0,0 +1,149 @@
...
+            sheet.activate();
+            
+            // Get cell A1 and write a number formatted as a date to cell A1.
+            const dateRange = sheet.getRange("A1");
+            dateRange.valuesAsJson = [
+              [
+                {
+                  type: Excel.CellValueType.formattedNumber,
+                  basicValue: 32889.0,
+                  numberFormat: "m/d/yyyy"
```
AlexJerabek, line 31: General feature feedback: This does not override the standard `Range.numberFormats`. If you click one of the formatting options in the Excel UI, it'll change the cell without the underlying JSON numberFormat changing. I found this confusing.
image

samples/excel/85-preview-apis/data-types-formatted-number.yaml
```diff
@@ -0,0 +1,149 @@
...
+          });
+        }
+
+        async function setFormattedNumberCurrency() {
+          await Excel.run(async (context) => {
+            // This method creates a formatted number data type,
+            // and sets the format of this data type as a currency.
+
+            // Get the active worksheet and a range on that sheet.
+            const sheet = context.workbook.worksheets.getActiveWorksheet();
```
AlexJerabek, line 45: If you're creating a "Sample" worksheet, you should get that sheet instead of the active one.

samples/excel/85-preview-apis/data-types-formatted-number.yaml
```diff
@@ -0,0 +1,149 @@
...
+            const currencyRange = sheet.getRange("A2");
+
+            // Write a number formatted as currency to cell A2.
+            currencyRange.valuesAsJson = [
```
AlexJerabek, line 49: What does it look like to set more than one cell at once? Is it really clunky or is there an optimization we could showcase?

samples/excel/85-preview-apis/data-types-formatted-number.yaml
```diff
@@ -0,0 +1,149 @@
...
+            // in cells A1 and A2 to the console.
+
+            // Get the active worksheet and two ranges on that sheet.
+            const sheet = context.workbook.worksheets.getActiveWorksheet();
+            const currencyRange = sheet.getRange("A2");
+            const dateRange = sheet.getRange("A1");
+
+            // Load the property of the ranges that contains data types.
+            currencyRange.load(["valuesAsJson"]);
+            dateRange.load(["valuesAsJson"]);
```
AlexJerabek, line 75: If you're loading a single property, you don't need the array.
```suggestion
            currencyRange.load("valuesAsJson");
            dateRange.load("valuesAsJson");
```

samples/excel/85-preview-apis/data-types-formatted-number.yaml
```diff
@@ -0,0 +1,149 @@
...
+          await Excel.run(async (context) => {
+            // This method prints information about data types 
+            // in cells A1 and A2 to the console.
+
+            // Get the active worksheet and two ranges on that sheet.
+            const sheet = context.workbook.worksheets.getActiveWorksheet();
+            const currencyRange = sheet.getRange("A2");
+            const dateRange = sheet.getRange("A1");
+
+            // Load the property of the ranges that contains data types.
```
AlexJerabek, line 73: This sentence reads a little awkwardly because on first pass it sounds like there's a grammar error in "ranges that contains data types". Consider reworking it to avoid potential confusion.
```suggestion
            // Load the data type property of the ranges.
```

samples/excel/85-preview-apis/data-types-formatted-number.yaml
```diff
@@ -0,0 +1,149 @@
...
+            currencyRange.load(["valuesAsJson"]);
+            dateRange.load(["valuesAsJson"]);
+            await context.sync();
+
+            // Access the data types property.
```
AlexJerabek, line 78: Nit: This strikes me as a comment that could be removed, as per our recent discussion with Ricky.

samples/excel/85-preview-apis/data-types-formatted-number.yaml
```diff
@@ -0,0 +1,149 @@
...
+            // Note: In a production add-in, you'd want to notify the user through your add-in's UI.
+            console.error(error);
+          }
+        }
+    language: typescript
+template:
+    content: |-
+        <section class="ms-font-m">
+          <p>This sample shows how to work with formatted number data types. The sample sets a number formatted as a date,
+            and a number formatted as currency. It also prints information about these data types to the console.</p>
```
AlexJerabek, line 110: Since the buttons do a good job explaining their functionality, consider trimming this description. Also, should formatted number be a singular data type?
```suggestion
          <p>This sample shows how to work with the formatted number data type.</p>
```

samples/excel/85-preview-apis/data-types-formatted-number.yaml
```diff
@@ -0,0 +1,149 @@
...
+        </section>
+        <section class="setup ms-font-m">
+          <h3>Try it out</h3>
+          <button id="set-formatted-number-date" class="ms-Button">
+              <span class="ms-Button-label">Set formatted date number</span>
```
AlexJerabek, line 115: Consider a shorter button name so it doesn't run the risk of wrapping.
```suggestion
              <span class="ms-Button-label">Set formatted date</span>
```

samples/excel/85-preview-apis/data-types-formatted-number.yaml
```diff
@@ -0,0 +1,149 @@
...
+          </button>
+          <button id="set-formatted-number-currency" class="ms-Button">
+              <span class="ms-Button-label">Set formatted currency number</span>
```
AlexJerabek, line 118: Consider a shorter button name so it doesn't run the risk of wrapping.
```suggestion
              <span class="ms-Button-label">Set formatted currency</span>
```

samples/excel/85-preview-apis/data-types-formatted-number.yaml
```diff
@@ -0,0 +1,149 @@
...
+          </button>
+          <button id="get-formatted-number" class="ms-Button">
+              <span class="ms-Button-label">Print data types to console</span>
```
AlexJerabek, line 121: Consider a shorter button name so it doesn't run the risk of wrapping.
```suggestion
              <span class="ms-Button-label">Log data types</span>
```

snippet-extractor-output/snippets.yaml
```diff
@@ -2500,6 +2500,139 @@
...
+    https://raw.githubusercontent.com/OfficeDev/office-js-snippets/prod/samples/excel/85-preview-apis/data-types-formatted-number.yaml
+
+    await Excel.run(async (context) => {
+      // This method creates a formatted number data type,
+      // and sets the format of this data type as a date.
+
+      // Create a worksheet called "Sample" and activate it.
+      context.workbook.worksheets.getItemOrNullObject("Sample").delete();
+      const sheet = context.workbook.worksheets.add("Sample");
+      sheet.activate();
```
AlexJerabek, line 2542: Moving the worksheet creation/deletion code will also take it out of the ref doc sample.

snippet-extractor-output/snippets.yaml
```diff
@@ -2500,6 +2500,139 @@
 
         await context.sync();
     });
+'Excel.FormattedNumberCellValue#basicType:member':
```
AlexJerabek, line 2503: Since this sample is repeated on the ref doc page, consider omitting one for `basicType`. That property is just read-only and not as interesting from a scenario perspective.

snippet-extractor-output/snippets.yaml
```diff
@@ -2500,6 +2500,139 @@
...
+            type: Excel.CellValueType.formattedNumber,
+            basicValue: 12.34,
+            numberFormat: "$* #,##0.00"
+          }
+        ]
+      ];
+
+      await context.sync();
+    });
+'Excel.FormattedNumberCellValue#type:member':
```
AlexJerabek, line 2609: The same comments about `basicValue` apply to `type` too.

snippet-extractor-output/snippets.yaml
```diff
@@ -2500,6 +2500,139 @@
...
+        [
+          {
+            type: Excel.CellValueType.formattedNumber,
+            basicValue: 32889.0,
+            numberFormat: "m/d/yyyy"
+          }
+        ]
+      ];
+      await context.sync();
+    });
```
AlexJerabek, line 2583: Consider omitting this sample for `numberFormat` and using the other one that mapped instead. That way, there's two and only to code samples on the ref doc page and they're both unique.

AlexJerabek (changes requested):
Looks good. I have a few feature questions and a suggestion for the setup flow.

snippet-extractor-output/snippets.yaml
alison-mk: Good point!

samples/excel/85-preview-apis/data-types-formatted-number.yaml
alison-mk: This is a great point. I will take this feedback to the feature crew.

samples/excel/85-preview-apis/data-types-formatted-number.yaml
alison-mk: The `numberFormat` property only controls the display of a value in the Excel UI. It doesn't change the underlying number value, even when the value changes dramatically through formatting as a date or currency. The discrepancy between the display and the underlying value can lead to errors during calculation and confusion for the end user. I think the short answer is that formatting with data types is safer, and once released to GA will be the recommended strategy for number formatting with add-ins. This is described in the conceptual docs. Do you think it should be addressed directly in the Script Lab sample, too?

samples/excel/85-preview-apis/data-types-formatted-number.yaml
alison-mk: I can look into this and, if an optimization strategy for multiple cells exists, include it in my next Script Lab sample for data types (web image is up next). Does that sound reasonable?

samples/excel/85-preview-apis/data-types-formatted-number.yaml
alison-mk: Maybe I'm misunderstanding the question. My answer assumed you were referring to [Excel.NumberFormatInfo class](https://docs.microsoft.com/en-us/javascript/api/excel/excel.numberformatinfo?view=excel-js-preview), which is a class and not a property (my mistake).

AlexJerabek: I was referring to [Range.numberFormat](https://docs.microsoft.com/en-us/javascript/api/excel/excel.range?view=excel-js-preview#excel-excel-range-numberformat-member). That's another way to set the format without changing the value.

samples/excel/85-preview-apis/data-types-formatted-number.yaml
AlexJerabek: Yeah, that's a good plan. Really, this is more of a question for the Data Types feature in general than for this particular sample.

samples/excel/85-preview-apis/data-types-formatted-number.yaml
```diff
@@ -0,0 +1,160 @@
...
+
+            const currencyValues = currencyRange.valuesAsJson[0][0];
+            const dateValues = dateRange.valuesAsJson[0][0];
+
+            // Print information about the data types to the console.
+            console.log("Date");
+            console.log("  Type: " + dateValues.type);
+            console.log("  Basic value: " + dateValues.basicValue);
+            console.log("  Basic type: " + dateValues.basicType);
+            console.log("  Number format: " + dateValues.numberFormat);
```
AlexJerabek, line 85: I'm getting compiler errors here, since `numberFormat` is only a property of `FormattedNumberCellValue`. Consider casting it like this:
```suggestion
            console.log("  Number format: " + (dateValues as Excel.FormattedNumberCellValue).numberFormat);
```

samples/excel/85-preview-apis/data-types-formatted-number.yaml
```diff
@@ -0,0 +1,160 @@
...
+
+            console.log("Currency");
+            console.log("  Type: " + currencyValues.type);
+            console.log("  Basic value: " + currencyValues.basicValue);
+            console.log("  Basic type: " + currencyValues.basicType);
+            console.log("  Number format: " + currencyValues.numberFormat);
```
AlexJerabek, line 91: Same comment as above.
```suggestion
            console.log("  Number format: " + (currencyValues as Excel.FormattedNumberCellValue).numberFormat);
```

AlexJerabek (approved):
Approving with a couple more suggestions.

samples/excel/85-preview-apis/data-types-formatted-number.yaml
alison-mk: Thanks! I thought that compiler error would be resolved once we fixed the type alias issue, but I was mistaken. It was unrelated.

PR #592, opened 2022-02-03
[Excel] (Data types) Add preview sample for web images

alison-mk:
[Internal work item.](https://office.visualstudio.com/OC/_workitems/edit/5763605)

@AlexJerabek I previously mentioned that I would include a strategy to update multiple cells in this web image data types sample, but after research I think that the formatted number value sample is better suited to showcasing how to update multiple cells. So, my plan is to update the FNV sample instead, after this web image sample is finalized.


samples/excel/85-preview-apis/data-types-web-image.yaml
```diff
@@ -0,0 +1,206 @@
...
+              console.log("Please enter an image URL.");
+              return;
+            }
+
+            // Insert web image into the active cell.
+            activeCell.valuesAsJson = [
+              [
+                {
+                  type: "WebImage",
+                  basicType: "Error",
```
AlexJerabek, line 40: Consider adding a /* comment */ explaining why "Error" is used here. Is it a best practice for types that have no basic value equivalent?

AlexJerabek, line 40: Can we use the enum instead of the string variant? I think the enums help teach the code better by showing the possibilities (instead of making it look like arbitrary strings can go here).

samples/excel/85-preview-apis/data-types-web-image.yaml
```diff
@@ -0,0 +1,206 @@
...
+            activeCell.load();
+            await context.sync();
+
+            if (!imageUrl) {
+              console.log("Please enter an image URL.");
+              return;
+            }
+
+            // Insert web image into the active cell.
+            activeCell.valuesAsJson = [
```
AlexJerabek, line 36: Instead of directly assigning the JSON, consider creating a WebImageCellValue variable as an intermediary step. That way, the type is clear and readers in the ref docs can know what to look up.

samples/excel/85-preview-apis/data-types-web-image.yaml
```diff
@@ -0,0 +1,206 @@
...
+            activeCell.load("valuesAsJson");
+            await context.sync();
+
+            // Get image URL from the active cell.
+            const values = activeCell.valuesAsJson;
+            const webImageData = values[0][0] as Excel.WebImageCellValue;
+            const webImageUrl = webImageData.address;
+
+            if (!webImageUrl) {
+              console.log("The selected cell is missing an image URL. Make sure to select a cell that contains an image.");
```
AlexJerabek, line 93: Consider shortening this a little.
```suggestion
              console.log("The selected cell is missing an image URL. Select a cell that contains an image.");
```

snippet-extractor-output/snippets.yaml
```diff
@@ -5129,6 +5129,82 @@
         textbox.textFrame.deleteText();
         await context.sync();
     });
+'Excel.WebImageCellValue#address:member':
```
AlexJerabek, line 5132: Is there value in having one of these two samples on the same page map to the `openImage` function instead? That way, you're showing both get and set functionality.

AlexJerabek (approved):
Approving with several suggestions.  I'm happy to take another look in the changes end up being more work.

samples/excel/85-preview-apis/data-types-web-image.yaml
alison-mk: Good point. Yes, it's a best practice.

alison-mk (commented):
Thanks for the review! Responding with a few questions.

samples/excel/85-preview-apis/data-types-web-image.yaml
```diff
@@ -0,0 +1,206 @@
...
+        </section>
+
+        <section class="setup ms-font-m">
+          <h3>Set up</h3>
+          <p>Add a new worksheet and then enter the URL and alt text for an image of your choice.</p>
+          <button id="setup" class="ms-Button">
+            <span class="ms-Button-label">Add worksheet</span>
+          </button>
+          <label for="url"><strong>Image URL:</strong> </label>
+          <input type="text" name="url" id="url" alt="Image url input field"><br>
```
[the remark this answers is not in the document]
AlexJerabek: There's a `.github/images` folder in this repo. Maybe putting an image there and referencing the published URL would be good (you'd have to add and test that separately). I don't think the current solution is bad, by any means, just thinking of ways to remove user friction.

samples/excel/85-preview-apis/data-types-web-image.yaml
```diff
@@ -0,0 +1,206 @@
...
+        <section class="samples ms-font-m">
+          <h3>Try it out</h3>
+          <p>Select the cell you want to insert the web image into, and then select the <strong>Insert image</strong> button.</p>
+          <button id="insert-image" class="ms-Button">
+            <span class="ms-Button-label">Insert image</span>
+          </button>
+
+          <p>Select the cell containing the web image that you want to view the details of, and then select the <strong>Retrieve image details</strong> button. The image details will display in the preceding <strong>Image URL</strong> and <strong>Alt text</strong> fields.</p>
+          <button id="retrieve-image-info" class="ms-Button">
+            <span class="ms-Button-label">Retrieve image details</span>
```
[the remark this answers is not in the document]
AlexJerabek: I mean the scenario where the user adds and image and clicks through the rest of the buttons. I didn't notice the fields change when first testing this.

samples/excel/85-preview-apis/data-types-web-image.yaml
AlexJerabek: I'm just biased against that string overload... I think the enum is easier to read and maintain from a user perspective. Your call either way.

alison-mk: I misspoke. `basicType` and `basicValue` are readonly properties. They can technically be written, but they have to be written as the values shown here. To simplify and avoid confusion, I will remove these two properties from the Script Lab sample.

samples/excel/85-preview-apis/data-types-web-image.yaml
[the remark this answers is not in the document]
alison-mk: I've updated the description before this button to better explain where the image details will display. I think it's sufficient, but if you feel strongly (or if we get feedback from customers) then I can adjust in the future.

samples/excel/85-preview-apis/data-types-web-image.yaml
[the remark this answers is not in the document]
alison-mk: I'll leave the image as-is for now, but if we get customer feedback that it's unclear then I will incorporate this suggestion.

alison-mk:
@AlexJerabek Thanks for the review and responses to my questions! I didn't make substantial changes to the sample (and provided explanations why not), so I'll merge instead of requesting that you take a second look.

PR #597, opened 2022-02-18
[Word] (content controls) How to work with onDeleted event

ElizabethSamuel-MSFT:
Fixes https://github.com/OfficeDev/office-js-docs-reference/issues/1145


playlists/word.yaml
```diff
@@ -191,6 +191,15 @@
   group: Scenarios
```
AlexJerabek, line 191: Instead of Scenarios, consider putting this under the "Content Controls" category.

samples/word/99-preview-apis/content-control-ondeleted-event.yaml
```diff
@@ -0,0 +1,161 @@
+id: word-content-control-ondeleted-event
+name: Content control onDeleted event
+description: Register and trigger onDeleted event on content controls.
```
AlexJerabek, line 3: Consider framing the name and description more around the scenario of deleting the controls.
```suggestion
name: Delete content controls
description: Register and trigger onDeleted event to track the removal of content controls.
```

samples/word/99-preview-apis/content-control-ondeleted-event.yaml
```diff
@@ -0,0 +1,161 @@
...
+            contentControls.load("items");
+            await context.sync();
+
+            // Register the onDeleted event handler on each content control.
+            if (contentControls.items.length === 0) {
+              console.log("There aren't any content controls in this document so can't register event handlers.");
+            } else {
+              for (let i = 0; i < contentControls.items.length; i++) {
+                contentControls.items[i].onDeleted.add(contentControlDeleted);
+                contentControls.items[i].track();
```
AlexJerabek, line 45: Why are these being manually tracked?

AlexJerabek (approved):
Looks good. Just a couple non-blocking questions and suggestions.

AlexJerabek:
@ElizabethSamuel-MSFT One more question: Did you update the word.xslx file in the snippet metadata so these go to the ref docs?

ElizabethSamuel-MSFT:
> @ElizabethSamuel-MSFT One more question: Did you update the word.xslx file in the snippet metadata so these go to the ref docs?

@AlexJerabek I'm actually doing that right now. I'll push an update soon.

playlists/word.yaml
ElizabethSamuel-MSFT: Actually, it's in the Preview APIs group. It'll be put under "Content Controls" if/when the onDeleted event GAs

samples/word/99-preview-apis/content-control-ondeleted-event.yaml
ElizabethSamuel-MSFT: This was recommended by the engineer. In this sample, the event doesn't fire (or least the handler isn't run) without that statement.

AlexJerabek: Hmm, I didn't encounter that issue (works fine for me without the tracking). If it is necessary, I'd recommend adding a comment explaining why it's needed.

AlexJerabek: Nevermind, I'm mistaken. I do think we should add a note explaining why that `track` statement is needed, since it deviates from the Excel event pattern.

PR #603, opened 2022-03-07
[tools] Adding top-level sample support for mapping

AlexJerabek:
This PR adds the ability to map a sample to the top-level of a reference doc page. In the example of `ChartType`, the given snippet lands in at the top of that enum page. This same logic will work for classes and interfaces.


AlexJerabek:
@ElizabethSamuel-MSFT Glad it was an easy fix :-)

PR #605, opened 2022-03-10
[Word] (content control) Add how to deregister onDeleted event

ElizabethSamuel-MSFT:


samples/word/99-preview-apis/content-control-ondeleted-event.yaml
```diff
@@ -76,6 +79,21 @@ script:
           });
         }
 
+        async function deregisterEventHandlers() {
```
AlexJerabek, line 82: This doesn't seem to work ☹️. I'm still seeing the event fire after pressing the deregister button.

image

ElizabethSamuel-MSFT: The delete will still happen but you shouldn't see the log from the handler ("ID of content control that was deleted:...") which you don't see here. However, something is still up since I don't see that log after the first item you deleted. I'll take another look.

ElizabethSamuel-MSFT: So it should look like this:

image

ElizabethSamuel-MSFT: As we discussed, this only applies to Windows

AlexJerabek (approved):
Looks good (on desktop 😉)

One suggestion that's separate from the last PR: can you add the phrase "Event detected" or something to the console.log statement in the event handler? That would have helped me realized the web problem earlier.

```
async function contentControlDeleted(event: Word.ContentControlEventArgs) {
            // Display the deleted content control's ID.
            console.log(`Event detected: ID of content control that was deleted: ${event.contentControl.id}`);
          });
        }
```

PR #614, opened 2022-03-23
Added Data Types: Error values and Entity from table samples

jakobpn:


alison-mk (approved):
Thanks for contributing, @jakobpn!

PR #620, opened 2022-04-05
[Excel] (Chart) Add ChartFill solidColor preview API

alison-mk:


alison-mk:
Closing PR. API has been removed from current release track.

PR #622, opened 2022-04-20
[Excel] (Chart) Add series data source preview API snippet

alison-mk:
This PR adds a snippet for the ChartSeries.getDimensionDataSourceString and ChartSeries.getDimensionDataSourceType preview APIs. It also corrects an error in excel.xlsx for the mapping of CellValueType.


samples/excel/85-preview-apis/chart-data-source.yaml
```diff
@@ -0,0 +1,129 @@
...
+            // Create a line chart based on data from Qtr1 from SalesTable.
+            let dataRange = sheet.getRange("A1:B7");
+            let chart = sheet.charts.add("Line", dataRange, "Auto");
+
+            // Position and style the chart.
+            chart.setPosition("A15", "E30");
+            chart.legend.position = "Right";
+            chart.legend.format.fill.setSolidColor("white");
+            chart.dataLabels.format.font.size = 15;
+            chart.dataLabels.format.font.color = "black";
```
AlexJerabek, line 68: I find this label covers up most of the chart. Can we remove these two lines?

samples/excel/85-preview-apis/chart-data-source.yaml
```diff
@@ -0,0 +1,129 @@
...
+              ["Spokes", 6005, 7634, 4589, 8765]
+            ]);
+            
+            sheet.getUsedRange().format.autofitColumns();
+            sheet.getUsedRange().format.autofitRows();
+            sheet.activate();
+
+            // Create a line chart based on data from Qtr1 from SalesTable.
+            let dataRange = sheet.getRange("A1:B7");
+            let chart = sheet.charts.add("Line", dataRange, "Auto");
```
AlexJerabek, line 61: Consider having the chart show part sales over time, instead of having each series be a quarter. This would have the following effects:
- The scenario is easier to logically follow.
- The chart will have multiple series, so it highlights the specifics of what the new API returns.
- It gives you some code to copy and paste as a sample for `ChartSeriesBy`.

```suggestion
            // Create a line chart based on data from SalesTable.
            let dataRange = sheet.getRange("A1:E7");
            let chart = sheet.charts.add("Line", dataRange, Excel.ChartSeriesBy.rows);
```

samples/excel/85-preview-apis/chart-data-source.yaml
```diff
@@ -0,0 +1,129 @@
...
+          <h3>Set up</h3>
+          <p>Add a product table and line chart to a Sample worksheet.</p>
+          <button id="setup" class="ms-Button">
+            <span class="ms-Button-label">Add sample data</span>
+          </button>
+        </section>
+
+        <section class="samples ms-font-m">
+          <h3>Try it out</h3>
+          <p>Log information to the console about the data source of the chart series <strong>Qtr1</strong>.</p>
```
AlexJerabek, line 100: Instead of this label, could we just have the button say "Log data source of X"?

samples/excel/85-preview-apis/chart-data-source.yaml
```diff
@@ -0,0 +1,129 @@
...
+    ExcelApi: BETA (PREVIEW ONLY)
+script:
+    content: |
+        $("#setup").click(() => tryCatch(setup));
+        $("#log-chart-series-source").click(() => tryCatch(logChartSeriesSource));
+
+        async function logChartSeriesSource() {
+          await Excel.run(async (context) => {
+            // This method retrieves the data source string and type values from
+            // the chart on the Sample worksheet.
```
AlexJerabek, line 16: Suggestion to clarify between the chart and chart series.
```suggestion
            // This method retrieves the data source information of a chart series in the Sample worksheet.
```

samples/excel/85-preview-apis/chart-data-source.yaml
```diff
@@ -0,0 +1,129 @@
...
+            const sheet = context.workbook.worksheets.getItem("Sample");
+
+            // Get the series collection from the first chart on the worksheet, and
+            // then get the first series in that collection.
```
AlexJerabek, line 20: Consider reframing this to be about the goal, rather than the step-by-step.
```suggestion
            // Get the first chart series from the first chart on the worksheet.
```

samples/excel/85-preview-apis/chart-data-source.yaml
```diff
@@ -0,0 +1,129 @@
...
+            const series = seriesCollection.getItemAt(0);
+
+            // Get the series data source string and type values. 
+            const dataSourceString = series.getDimensionDataSourceString("Values");
+            const dataSourceType = series.getDimensionDataSourceType("Values");
+            await context.sync();
+            
+            // Log the data source information to the console.
+            console.log("Qtr1 data source string: " + dataSourceString.value);
+            console.log("Qtr1 data source type: " + dataSourceType.value);
```
AlexJerabek, line 31: Since you have the chart series, consider replacing Qtr1 with `series.name` (be sure to add a load statement before the `context.sync`).

AlexJerabek (changes requested):
Looks good. I left a suggestion about changing the chart, which would have a few side effects worth rechecking.

samples/excel/85-preview-apis/chart-data-source.yaml
alison-mk: Thank you for the review! Great feedback. I've incorporated all of your changes except this one about the button text. I think the description above the button is helpful and I want to keep it. 

The chart now displays 6 series, and the button text would be "Log data source of Frames". Since "Frames" appears in two places on the worksheet, I want to clarify that this is logging source info for the chart series Frames, not the table Frames, which would make the button text "Log data source of chart series Frames" — and that's too long for a button.

AlexJerabek (approved):
Thanks for making those changes!

PR #626, opened 2022-04-27
[Excel] (Data types) Add entity value data provider attribution snippet

alison-mk:
@AlexJerabek The `products.sourceAddress` image URLs in this snippet won't work until this PR is merged to main. For testing, you could switch them out for image URLs to this PR instead (such as https://github.com/OfficeDev/office-js-snippets/blob/945c1b9bc085ad5fa804793ca2cd46f213524708/.github/images/microsoft-logo.png?raw=true


samples/excel/85-preview-apis/data-types-entity-attribution.yaml
```diff
@@ -0,0 +1,235 @@
...
+            targetAddress: "http://microsoft.com"
+          },
+          {
+            productID: 3,
+            productName: "Aniseed Syrup",
+            quantityPerUnit: "12 - 550 ml bottles",
+            unitPrice: 10,
+            providerName: "MSN",
+            sourceAddress: "https://github.com/OfficeDev/office-js-snippets/blob/main/.github/images/msn-logo.png?raw=true",
+            targetAddress: "http://xbox.com"
```
AlexJerabek, line 169: I think this address should be for MSN.

samples/excel/85-preview-apis/data-types-entity-attribution.yaml
```diff
@@ -0,0 +1,235 @@
...
+          },
+          {
+            productID: 4,
+            productName: "Chef Anton's Cajun Seasoning",
+            quantityPerUnit: "48 - 6 oz jars",
+            unitPrice: 22,
+            providerName: "Xbox",
```
AlexJerabek, line 176: Consider using Bing instead of Xbox, if only because the idea of Xbox dealing with seasoning is a bit jarring (get it? because it's sold in jars? 🤣)

samples/excel/85-preview-apis/data-types-entity-attribution.yaml
```diff
@@ -0,0 +1,235 @@
...
+                    collapsed: false,
+                    properties: ["Quantity Per Unit", "Unit Price"]
+                  }
+                ]
+              }
+            },
+            provider: {
+              description: product.providerName, // Name of the data provider. Displays as a tooltip when hovering over the logo. Also displays as a fallback if the source address for the image is broken.
+              logoSourceAddress: product.sourceAddress, // Source URL of the logo to display.
+              logoTargetAddress: product.targetAddress // Destination URL that the logo navigates to when clicked.
```
AlexJerabek, line 95: Since this block is what the sample is highlighting, is it possible to add this part as a separate button press? So, the workflow would be: 
- Add the table
- Add the entity values
- Add the provider data

That way, the difference between the two states would be more obvious.

samples/excel/85-preview-apis/data-types-entity-attribution.yaml
```diff
@@ -0,0 +1,235 @@
+order: 7
+id: excel-data-types-entity-attribution
+name: 'Data types: Entity value attribution properties'
+description: This sample shows how to set entity values with data provider attributions in the card layout.
```
AlexJerabek, line 4: Consider mentioning the featured properties first.
```suggestion
description: This sample shows how to set data provider attributions on entity values in the card layout.
```

AlexJerabek (commented):
Looks good @alison-mk. I have a consideration for a slight structural change. Let me know what you think.

samples/excel/85-preview-apis/data-types-entity-attribution.yaml
alison-mk: Lol, agreed, but will leave as Xbox based on offline conversation about logo use case guidance.

AlexJerabek (approved):

samples/excel/85-preview-apis/data-types-entity-attribution.yaml
alison-mk: Based on offline conversation, I'll leave the workflow as is, and I added a sentence about the "data aggregator" scenario in the HTML.

PR #635, opened 2022-05-07
[Excel] (Data types) Add preview sample for entity value references

alison-mk:
This PR adds a snippet showing how to create entity value data types that have references to other entity value data types.


samples/excel/85-preview-apis/data-types-references.yaml
```diff
@@ -0,0 +1,337 @@
...
+
+    // Collect all the references to include in the employee entity.
+    // A reference is a link to another entity value outside the original entity.
+    function collectReferences(employee) {
+      const references: number[] = [];
+      collectReferencesM(employee, references);
+      return references;
+    }
+
+    function collectReferencesM(employee, references: number[]) {
```
alison-mk, line 77: @jakobpn What does the "M" stand for in the method name `collectReferencesM`?

jakobpn: M stands for "Mark". So just to give it a different name from collectReferences.

playlists-prod/excel.yaml
```diff
@@ -1201,6 +1201,18 @@
   group: Preview APIs
   api_set:
     ExcelApi: BETA (PREVIEW ONLY)
+- id: excel-data-types-references
+  name: 'Data types: Entity values with references'
+  fileName: data-types-references.yaml
+  description: >-
+    This sample shows how to create entity values with references to other
+    entity values. An entity value is a container for data, and this container
+    can have references (or links) to other entity values on the same worksheet.
```
jakobpn, line 1210: The "on the same worksheet" is not correct. It is not possible for an entity value to have references to other entities on the worksheet. The references can only be within an entity value.

jakobpn: Also, I don't think of "references" link, but rather a pointer to an entity value.

samples/excel/85-preview-apis/data-types-references.yaml
```diff
@@ -0,0 +1,339 @@
...
+            firstName: "Anne",
+            title: "Sales Representative",
+            reportsTo: 4
+          }
+        ];
+    language: typescript
+template:
+    content: |
+        <section class="ms-font-m">
+          <p>This sample shows how to create entity values with references to other entity values. An entity value is a container for data, and this container can have references (or links) to other entity values on the same worksheet.</p>
```
jakobpn, line 297: "On same worksheet" is incorrect. See explanation in comment above.

jakobpn (changes requested):
Thx for the opportunity to review. I added a few comments.

samples/excel/85-preview-apis/data-types-references.yaml
```diff
@@ -0,0 +1,339 @@
...
+          </button>
+        </section>
+        <section class="samples ms-font-m">
+          <h3>Try it out</h3>
+          <button id="add-entities-to-table" class="ms-Button">
+            <span class="ms-Button-label">Add entity values</span>
+          </button>
+          <p>To see the entity value, click the icon to the left of the title in the <strong>Employee</strong> column after
+            selecting <strong>Add entity values</strong>.</p>
+          <p>Inside the card modal window, select the <strong>Extract to grid</strong> icon to the right of the <strong>Manager</strong> field or the <strong>Direct Reports</strong> field to the grid to display the referenced entities.</p>
```
AlexJerabek, line 312:
```suggestion
          <p>Inside the card modal window, select the <strong>Extract to grid</strong> icon to the right of the <strong>Manager</strong> field or the <strong>Direct Reports</strong> field to display the referenced entities.</p>
```

AlexJerabek, line 312: Consider changing "display the referenced entities" to "add the referenced entities to the table" or something. "Display" makes me think it'll happen in the control or the task pane.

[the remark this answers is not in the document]
alison-mk: The "#FIELD!" error is for an employee that doesn't have a manager (the top level employee in the hierarchy).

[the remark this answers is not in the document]
AlexJerabek: Right, but is that controllable by the add-in? It seems like it'd be better to say "None" or "N/A" or something.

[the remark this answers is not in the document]
jakobpn: This is the "as designed" behavior of data types. We have this for Stocks and Geography, for example. This has the benefit that the Excel user can decide what to do when this error occurs. Users typically add IFERROR(...) around the field reference to handle this.

samples/excel/85-preview-apis/data-types-references.yaml
```diff
@@ -0,0 +1,339 @@
...
+
+        // Create entities from employee properties.  
+        function makeEmployeeEntity(lastName: string, firstName: string, employee: any, references: number[]): Excel.EntityCellValue {
+          const entity: Excel.EntityCellValue = {
+            type: Excel.CellValueType.entity,
+            text: `${firstName} ${lastName}`,
+            properties: {
+              "Employee ID": {
+                type: Excel.CellValueType.string,
+                basicValue: employee.employeeID.toString() || ""
```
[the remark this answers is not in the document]
jakobpn: Sure. It can be deleted.

samples/excel/85-preview-apis/data-types-references.yaml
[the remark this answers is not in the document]
alison-mk: @jakobpn Can you clarify — Are you suggesting that we leave this as a #FIELD! error in the code snippet?

AlexJerabek (approved):

jakobpn (approved):
Thanks for finalizing this sample, Alison!

PR #646, opened 2022-06-01
 [Excel] (Data types) Add preview sample for custom functions integration

alison-mk:


samples/excel/85-preview-apis/data-types-custom-functions.yaml
```diff
@@ -0,0 +1,334 @@
...
+id: excel-data-types-custom-functions
+name: 'Data types: Custom functions'
+description: This sample shows how to write custom functions that return entity value data types.
+host: EXCEL
+api_set:
+    ExcelApi: BETA (PREVIEW ONLY)
+script:
+    content: |
+        /**
+         * Search for products that match a given substring. Try =SCRIPTLAB.DATATYPESCUSTOMFUNCTIONS.PRODUCTSEARCH("syrup").
```
AlexJerabek, line 11: Consider adding a test case that uses `lookAt`.

samples/excel/85-preview-apis/data-types-custom-functions.yaml
```diff
@@ -0,0 +1,334 @@
...
+         * @customfunction
+         * @param search The string to search for in the sample JSON data.
+         * @param [lookAt] Define whether the search should be a match of the whole product name or part of the product name. If omitted, set to lookAt = "Part".
+         * @return {Promise<any[][]>} Search results as one or more data type entity values.
+         */
+        async function productSearch(query: string, lookAt?: string): Promise<any[][]> {
```
AlexJerabek, line 17: It's not clear to me what else `lookAt` could be. Should this just be a boolean value like `completeMatch` in [SearchCriteria](https://docs.microsoft.com/en-us/javascript/api/excel/excel.searchcriteria?view=excel-js-preview#excel-excel-searchcriteria-completematch-member)?

samples/excel/85-preview-apis/data-types-custom-functions.yaml
```diff
@@ -0,0 +1,334 @@
...
+            // Create product entities for each of the products returned in the search result.
+            const entities = searchResult.map((product) => [makeProductEntity(product)]);
+            return entities;
+          } catch (error) {
+            console.error(error);
+          }
+        }
+
+        // Helper method to create entities from product properties.
+        function makeProductEntity(product?: any) {
```
AlexJerabek, line 50: Are these helper methods recreating the JSON every time? This might be out of the scope of this PR/sample, but if they are, we should consider reworking this to only create the JSON on the first call and reference the stored data every other time (Singleton pattern).

AlexJerabek (approved):
Looks good. I have a couple small comments for immediate consideration and one longer term one.

samples/excel/85-preview-apis/data-types-custom-functions.yaml
alison-mk: @jakobpn Can you weigh in on this?

samples/excel/85-preview-apis/data-types-custom-functions.yaml
alison-mk: Interesting point, definitely open to considering this. @jakobpn Can you answer this question? --> 

> Are these helper methods recreating the JSON every time?

jakobpn: Yes, I think that is demonstrating how a real solution that would get the JSON from a service and then construct the entity value for each result. Plus, I think it beyond the purpose of the sample to show how to code a cache mechanism.

samples/excel/85-preview-apis/data-types-custom-functions.yaml
jakobpn: Agree. That is not particularly well documented in the signature of the function. I would suggest adding the 2 possible options to the @param comment for lookAt. I think I called this "lookAt" to be similar to the find methods in the Office JS API.

alison-mk: Thanks for the feedback. I'll change this param to a boolean called completeMatch so that the options (true/false) are more obvious.

samples/excel/85-preview-apis/data-types-custom-functions.yaml
alison-mk: Ok, thanks! I agree with Jakob and will close this issue. Reopen if you'd like further discussion.

PR #649, opened 2022-06-07
[Outlook] (EWS) Gets appointment's iCalUId as the organizer

samantharamon:


samples/outlook/85-tokens-and-service-calls/get-icaluid-as-organizer.yaml
```diff
@@ -0,0 +1,85 @@
...
+api_set:
+    Mailbox: '1.3'
+script:
+    content: |
+        $("#run").click(run);
+
+        function run() {
+          Office.context.mailbox.item.saveAsync((result) => {
+            if (result.status === Office.AsyncResultStatus.Failed) {
+              console.log(result.error.message);
```
ElizabethSamuel-MSFT, line 15: Consider using console.error so it will log it with error coloring and such...here and other instances
```suggestion
              console.error(result.error.message);
```

ElizabethSamuel-MSFT (approved):

PR #654, opened 2022-06-15
[excel] (PivotTable) Refresh sample

AlexJerabek:
Partially based on https://github.com/OfficeDev/office-js-docs-pr/issues/3464.

This sample shows how to refresh a PivotTable. It also highlights that PivotTables need table sources to pick up row additions (range sources don't adapt to changes in the range size).


view/excel.json
```diff
@@ -56,10 +56,11 @@
...
   "excel-pivottables-get-pivottables": "https://raw.githubusercontent.com/OfficeDev/office-js-snippets/main/samples/excel/38-pivottable/pivottable-get-pivottables.yaml",
-  "excel-pivottable-slicer": "https://raw.githubusercontent.com/OfficeDev/office-js-snippets/main/samples/excel/38-pivottable/pivottable-slicer.yaml",
   "excel-pivottables-pivotfilters": "https://raw.githubusercontent.com/OfficeDev/office-js-snippets/main/samples/excel/38-pivottable/pivottable-pivotfilters.yaml",
   "excel-pivottable-pivotlayout": "https://raw.githubusercontent.com/OfficeDev/office-js-snippets/main/samples/excel/38-pivottable/pivottable-pivotlayout.yaml",
   "excel-pivottable-data-source": "https://raw.githubusercontent.com/OfficeDev/office-js-snippets/main/samples/excel/38-pivottable/pivottable-source-data.yaml",
+  "excel-pivottable-refresh": "https://raw.githubusercontent.com/OfficeDev/office-js-snippets/main/samples/excel/38-pivottable/pivottable-refresh.yaml",
+  "excel-pivottable-slicer": "https://raw.githubusercontent.com/OfficeDev/office-js-snippets/main/samples/excel/38-pivottable/pivottable-slicer.yaml",
```
alison-mk, line 63: Out of curiosity, why move Slicer away from PivotFilters in the list?

samples/excel/38-pivottable/pivottable-refresh.yaml
```diff
@@ -0,0 +1,139 @@
...
+        async function addTableRow() {
+          await Excel.run(async (context) => {
+            // Add a row to the PivotTable's source table.
+            const dataTable = context.workbook.tables.getItem("DataTable");
+            dataTable.rows.add(null, [["G Farms", "Mango", "Organic", 200, 1000]]);
+            await context.sync();
+          });
+        }
+
+        async function refreshPivotTable() {
```
alison-mk, line 23: It seems like the point of this snippet is to showcase the refresh() method. Should the refreshPivotPivotTable function be listed first, before addTableRow? 

Also, our recent approach has been to include a sort of thesis statement in the primary/main function in a snippet (the method that is extracted into ref docs), such as: 

"This method shows how to refresh the "Farm Sales" PivotTable to see if new data has been added to the source."

alison-mk (approved):
Thanks for adding this! Approving with a couple comments.

view/excel.json
AlexJerabek: I was sorting them alphabetically, but I get the grouping. Considering there's another "Filters and summaries" snippet that's further up, I think I'd prefer to leave as is in this PR. That said, it would probably be good to reconcile the three PivotTable filtering mechanisms introduced across three requirement sets.

PR #669, opened 2022-07-22
[Outlook] Add MailboxEnums to enum names

ElizabethSamuel-MSFT:
@samantharamon I realized that the samples didn't get picked up in the -reference repo because we're missing part of the UID.


samantharamon (approved):
Thanks for catching this. Looks good to me!

PR #684, opened 2022-08-05
PPT app-specific basic samples and reorganize the PPT snippets folders

Rick-Kirkham:


AlexJerabek (approved):
Looks good to me! We'll have to double check once the playlist is in Script Lab and easily viewable, but everything seems to be in order.

PR #687, opened 2022-08-11
[excel] (Custom Functions) Allow for mapping for custom functions and their JSDoc

AlexJerabek:
This PR makes two changes:
- The snippet metadata files now specify a package, allowing for mappings to different namespaces (like CustomFunctions).
- The extracted snippet now includes the preceding JSDoc. This PR does not take full advantage of the new capability.

The downstream effects on office-js-docs-reference have not been fully tested.


ElizabethSamuel-MSFT (approved):

alison-mk (approved):
Thanks for doing this!

PR #689, opened 2022-08-12
[excel] (Custom Functions) Change StreamingInvocation.setResult to a property mapping

AlexJerabek:
I mistakenly assumed `setResult` was a method instead of a property.


ElizabethSamuel-MSFT (approved):
Confirmed that's the case in the CustomFunctions.StreamingInvocation documentation.

PR #697, opened 2022-08-31
[Outlook] (attachments) Clarify how to get attachments in read and compose mode

samantharamon:


ElizabethSamuel-MSFT (approved):
One small typo. And thinking we should also map handleAttachmentsCallback to the various getAttachmentContentAsync? But will it be clear that it's the callback function?

samantharamon:
> One small typo. And thinking we should also map handleAttachmentsCallback to the various getAttachmentContentAsync? But will it be clear that it's the callback function?

Mapping `handleAttachmentsCallback` may cause a bit of confusion, especially if the user sees that first before the relevant snippet. My thought is that we can safely exclude the callback function and if the user wants to test/run the sample, they can view the code via the link provided. However, to clarify the snippet and mention what the callback does, how about we add a comment above the `getAttachmentContentAsync` call? Something like, "// Log the attachment type and its contents to the console." Let me know your thoughts on this. Happy to discuss further.

ElizabethSamuel-MSFT:
> > One small typo. And thinking we should also map handleAttachmentsCallback to the various getAttachmentContentAsync? But will it be clear that it's the callback function?
> 
> Mapping `handleAttachmentsCallback` may cause a bit of confusion, especially if the user sees that first before the relevant snippet. My thought is that we can safely exclude the callback function and if the user wants to test/run the sample, they can view the code via the link provided. However, to clarify the snippet and mention what the callback does, how about we add a comment above the `getAttachmentContentAsync` call? Something like, "// Log the attachment type and its contents to the console." Let me know your thoughts on this. Happy to discuss further.

I think the snippets are listed in the order they appear in the spreadsheet. That said, I think your idea of placing a comment above the getAttachmentContentAsync statement is reasonable and also helps document the snippet itself so I'm for it.

samantharamon:
> > > One small typo. And thinking we should also map handleAttachmentsCallback to the various getAttachmentContentAsync? But will it be clear that it's the callback function?
> > 
> > 
> > Mapping `handleAttachmentsCallback` may cause a bit of confusion, especially if the user sees that first before the relevant snippet. My thought is that we can safely exclude the callback function and if the user wants to test/run the sample, they can view the code via the link provided. However, to clarify the snippet and mention what the callback does, how about we add a comment above the `getAttachmentContentAsync` call? Something like, "// Log the attachment type and its contents to the console." Let me know your thoughts on this. Happy to discuss further.
> 
> I think the snippets are listed in the order they appear in the spreadsheet. That said, I think your idea of placing a comment above the getAttachmentContentAsync statement is reasonable and also helps document the snippet itself so I'm for it.

Thanks for the feedback. I'll make that change now.

ElizabethSamuel-MSFT (approved):

PR #709, opened 2022-10-07
[Word] (Settings) Add sample for SettingCollection

ElizabethSamuel-MSFT:


samples/word/50-document/manage-settings.yaml
```diff
@@ -0,0 +1,133 @@
...
+            <p>Add a setting</p>
+            <p>
+                <div class="ms-TextField">
+                    <label class="msLabel">Key:</label>
+                    <input id="key" class="msTextField-field" type="text">
+                </div>
+                    <div class="ms-TextField">
+                        <label class="msLabel">Value:</label>
+                        <input id="value" class="msTextField-field" type="text">
+                </div>
```
AlexJerabek, line 94: Small nit, but is it possible to get these text input fields to line up?

image

samples/word/50-document/manage-settings.yaml
```diff
@@ -0,0 +1,133 @@
...
+    WordApi: '1.4'
+script:
+    content: |
+        $("#create-setting").click(() => tryCatch(createSetting));
+        $("#get-settings").click(() => tryCatch(getSettings));
+        $("#delete-all-settings").click(() => tryCatch(deleteAllSettings));
+
+        async function createSetting() {
+          await Word.run(async (context) => {
+            let key = $("#key")
```
AlexJerabek, line 16: Most of the `let`s should be `const`s in this snippet, since they don't change after initialization.

AlexJerabek, line 16: Let's add some "thesis statement" comments to the starts of each of these methods. They'll help the reader both here and in the reference docs once these snippets are plucked over.

samples/word/50-document/manage-settings.yaml
```diff
@@ -0,0 +1,133 @@
...
+              .val()
+              .toString();
+            let value = $("#value")
+              .val()
+              .toString();
+            const settings = context.document.settings;
+            let setting = settings.add(key, value);
```
AlexJerabek, line 23: The snippet throws an error for an empty key. Consider adding a check, unless you feel it detracts from the sample too much.

samples/word/50-document/manage-settings.yaml
```diff
@@ -0,0 +1,133 @@
+order: 6
+id: word-document-manage-settings
+name: Manage settings
+description: 'This sample shows how to add, get, and delete settings in the file.'
```
AlexJerabek, line 4: Is it worth defining "settings"? To me, settings are like configurations (zoom, ribbon locking, etc). What's the user scenario?

samples/word/50-document/manage-settings.yaml
AlexJerabek: Might be worth noting for the add that adding a setting to an existing key replaces the existing value.

samples/word/50-document/manage-settings.yaml
```diff
@@ -0,0 +1,133 @@
...
+            const allSettings = [];
+            for (let i = 0; i < settings.items.length; i++) {
+              let setting = settings.items[i];
+              setting.load();
+              allSettings.push(setting);
+            }
+
+            await context.sync();
+
+            if (allSettings.length <= 0) {
```
AlexJerabek, line 47: Nit since the length can't be less than 0 (and we shouldn't imply otherwise).
```suggestion
            if (allSettings.length == 0) {
```

AlexJerabek (changes requested):
Some minor requests, but looks good overall.

samples/word/50-document/manage-settings.yaml
```diff
@@ -0,0 +1,133 @@
+order: 6
+id: word-document-manage-settings
+name: Manage settings
+description: 'This sample shows how to add, get, and delete settings in the file.'
+host: WORD
+api_set:
+    WordApi: '1.4'
+script:
+    content: |
+        $("#create-setting").click(() => tryCatch(createSetting));
```
jipyua, line 10: can we change the function name to addSetting instead of createSetting?

samples/word/50-document/manage-settings.yaml
```diff
@@ -0,0 +1,133 @@
...
+          await Word.run(async (context) => {
+            let settings = context.document.settings;
+            settings.load("items");
+            await context.sync();
+
+            const allSettings = [];
+            for (let i = 0; i < settings.items.length; i++) {
+              let setting = settings.items[i];
+              setting.load();
+              allSettings.push(setting);
```
jipyua, line 42: seems no need to define this allSettings variable, directly console.log() each setting value is enough.

jipyua (approved):

PR #719, opened 2022-10-18
[Word] Update existing sample to show BodyType

ElizabethSamuel-MSFT:


snippet-extractor-output/snippets.yaml
```diff
@@ -13357,6 +13357,50 @@
...
+
+      const referenceNumber = $("#input-reference").val();
+      const mark = (referenceNumber as number) - 1;
+      const item = footnotes.items[mark];
+      console.log(`Note type of footnote ${referenceNumber}: ${item.type}`);
+
+      item.body.load("type");
+      await context.sync();
+
+      console.log(`Body type of note: ${item.body.type}`);
```
alison-mk, line 13402: For me, the console.log on line 13397 in this file and the console.log on 13402 return the same thing: "Footnote"

image

Should the `item.type` and the `item.body.type` be different?

snippet-extractor-output/snippets.yaml
```diff
@@ -13357,6 +13357,50 @@
...
+    });
+'Word.BodyType:enum':
+  - >-
+    // Link to full sample:
+    https://raw.githubusercontent.com/OfficeDev/office-js-snippets/prod/samples/word/50-document/manage-footnotes.yaml
+
+    // Get the referenced note's type, which is footnote.
```
alison-mk, line 13387: Perhaps this as an explanation: 
```suggestion
    // Get the referenced note's item type and body type, which are both footnote.
```

alison-mk (approved):
Approving with a question and a suggestion.

PR #722, opened 2022-10-18
[Word] Add snippet that shows how to get details about table cell pad…

ElizabethSamuel-MSFT:
…ding


AlexJerabek:
@ElizabethSamuel-MSFT What's the value of separating this sample from the table alignment sample? Could we instead have a single sample that showcases "table details"? That way, developers can see the similarities between these APIs and learn the application-specific patterns.

PR #749, opened 2023-01-25
[excel] Remove duplicate Excel snippets

AlexJerabek:
This PR looks like a lot, but it's because the Excel mapping spreadsheet was sorted. All that's happening is the removal of a couple mappings sending duplicate snippets to the same API.


package-lock.json
```diff
@@ -1,2939 +0,0 @@
-{
```
alison-mk, line 1: What is the motivation for deleting this file?

alison-mk (commented):
The snippets update looks good to me, but I do have a question about the package-lock.json file change in this PR.

package-lock.json
AlexJerabek: A package-lock file implies npm is used, instead of yarn. Since this repo uses yarn, it should only have a yarn.lock file. There's a warning to that effect when you call `yarn start`

alison-mk (approved):

PR #767, opened 2023-02-28
Adding closeAsync snippet sample

ndeleuze:


samples/outlook/99-preview-apis/close-async.yaml
```diff
@@ -0,0 +1,55 @@
...
+description: Close the current item and discard the changes when specified.
+host: OUTLOOK
+api_set:
+    Mailbox: preview
+script:
+    content: |
+        $("#closeAsync").click(closeAsync);
+
+        function closeAsync() {
+          // This snippet closes and discards the current item. This api call works on new compose messages and existing drafts when the optional parameter, discardItem, is set to true. When set to false, on new compose messages, the user is prompted to either save or discard the changes or cancel the close operation
```
samantharamon, line 13:
```suggestion
          // This snippet closes the current message being composed and discards any unsaved changes. The API call works on new messages in compose, inline replies, and existing drafts when the optional parameter, discardItem, is set to true. When discardItem is set to false on new messages in compose, the user is prompted to either save or discard the changes or cancel the close operation.
```

samantharamon (changes requested):
Thanks for creating this snippet, @ndeleuze! Added a few comments for your review to make the format consistent with that of existing snippets.

samples/outlook/99-preview-apis/close-async.yaml
ndeleuze: Hey @samantharamon I didn't see the prompt to Commit suggestion, so I did it manually (turns out I wasn't signed into Github😅). Let me know if everything looks good, thanks!

samantharamon (approved):

PR #805, opened 2023-07-13
[Word] (list) Log list properties

ElizabethSamuel-MSFT:


samantharamon (approved):
Approving with a suggestion to consider adding the `getListProps` snippet to [Word.List.levelExistences](https://learn.microsoft.com/en-us/javascript/api/word/word.list?view=word-js-preview#word-word-list-levelexistences-member).

PR #838, opened 2023-12-25
add content-genration scenario sample

YijunMS:
The new sample is part of the scenario experiment for Word JS developers. I changed the order of existing scenario samples to make the new sample more conspicuous. It would be good if we can move the whole "scenarios" folder to an earlier position, but I am not sure if it is safe to do that.

### DIFF
--- a/playlists-prod/word.yaml
+++ b/playlists-prod/word.yaml
@@ -436,6 +436,17 @@
   group: Document
   api_set:
     WordApi: '1.5'
+- id: word-scenarios-content-generation
+  name: Content Generation
+  fileName: content-generation.yaml
+  description: >-
+    Insert predefined or AI-generated content to different locations of the
+    document.
+  rawUrl: >-
+    https://raw.githubusercontent.com/OfficeDev/office-js-snippets/prod/samples/word/90-scenarios/content-generation.yaml
+  group: Scenarios
+  api_set:
+    WordApi: '1.5'
 - id: word-scenarios-doc-assembly
   name: Document assembly
   fileName: doc-assembly.yaml
--- a/playlists/word.yaml
+++ b/playlists/word.yaml
@@ -436,6 +436,17 @@
   group: Document
   api_set:
     WordApi: '1.5'
+- id: word-scenarios-content-generation
+  name: Content Generation
+  fileName: content-generation.yaml
+  description: >-
+    Insert predefined or AI-generated content to different locations of the
+    document.
+  rawUrl: >-
+    https://raw.githubusercontent.com/OfficeDev/office-js-snippets/main/samples/word/90-scenarios/content-generation.yaml
+  group: Scenarios
+  api_set:
+    WordApi: '1.5'
 - id: word-scenarios-doc-assembly
   name: Document assembly
   fileName: doc-assembly.yaml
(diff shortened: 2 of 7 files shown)


xiruatms:
@AlexJerabek , will you please help to review when available? Thanks.

samples/word/05-scenarios/content-generation.yaml
```diff
@@ -0,0 +1,507 @@
...
+    language: typescript
+template:
```
AlexJerabek, line 258: The task pane looks very different from other Script Lab samples. Please use the same layout and style as other samples.

samples/word/05-scenarios/content-generation.yaml
```diff
@@ -0,0 +1,507 @@
...
+        };
+
+        const predefinedData = {
+          title: "This is a predefined title",
+          comment: "This is a predefined comment",
+          citation:
+            "The Power of Storytelling in Education.\" TED, www.ted.com/talks/amanda_stevens_graham_the_power_of_storytelling_in_education."
+        };
+
+        function checkazureAIParams(): boolean {
```
AlexJerabek, line 41: These pop-ups get cutoff in the task pane. Is there a way to prevent that?

image

samples/word/05-scenarios/content-generation.yaml
```diff
@@ -0,0 +1,507 @@
...
+                        <div class="back">
+                            <div class="cursor" id="back">
+                                <span>Back</span>
+                            </div>
+                        </div>
+                        <div class="main_func">
+                            <div class="dropdown">
+                                <button class="dropdown_list">Add a title</button>
+                                <div class="dropdown-options">
+                                    <a href="#" id="title-predefined">Add a predefined title</a>
```
AlexJerabek, line 276: Let's not include "predefined" content. We have plenty of samples that show that. Let's focus on the AI part.

samples/word/05-scenarios/content-generation.yaml
```diff
@@ -0,0 +1,507 @@
...
+        const predefinedData = {
+          title: "This is a predefined title",
+          comment: "This is a predefined comment",
+          citation:
+            "The Power of Storytelling in Education.\" TED, www.ted.com/talks/amanda_stevens_graham_the_power_of_storytelling_in_education."
+        };
+
+        function checkazureAIParams(): boolean {
+          if (!azureAIParams.endpoint) {
+            azureAIParams.endpoint = prompt("Input your Azure AI service endpoint:", azureAIParams.endpoint);
```
AlexJerabek, line 43: What should I put here? Do we expect customers to have an Azure service ready to go? How is that communicated to the customer?

AlexJerabek (changes requested):
It seems like the user needs an Azure service to use this sample. Without clear instructions on how to set that up, customers will struggle to complete the prompts and use the sample.

If users need to create such a service, we should move this from Script Lab to the Code Samples repo, where we can have a readme. That might be a better place for a scenario-driven sample anyway. It's much easier to link to those samples than a Script Lab sample.

Tagging @davidchesnut for awareness.

samples/word/05-scenarios/content-generation.yaml
tech-jun-jones: function getData() {
    Office.context.document.getSelectedDataAsync(Office.CoercionType.Table, function(asyncResult) {
        if (asyncResult.status == Office.AsyncResultStatus.Failed) {
            write(asyncResult.error.message);
        }
        else {
            write(asyncResult.value);
        }
    });
}
// Function that writes to a div with id='message' on the page.
function write(message){
    document.getElementById('message').innerText += message; 
}

PR #841, opened 2024-01-30
[Word] (insert external doc) Add function to apply settings

ElizabethSamuel-MSFT:
I'll do metadata mapping after.

### DIFF
--- a/samples/word/50-document/insert-external-document.yaml
+++ b/samples/word/50-document/insert-external-document.yaml
@@ -1,14 +1,15 @@
 order: 3
 id: word-document-insert-external-document
 name: Insert an external document
-description: Inserts the text of an external document into the current document.
+description: 'Inserts the content (with or without settings) of an external document into the current document. Settings include formatting, change-tracking mode, custom properties, and XML parts.'
 host: WORD
 api_set:
-    WordApi: '1.3'
+    WordApi: '1.6'
 script:
     content: |
         $("#file").change(getBase64);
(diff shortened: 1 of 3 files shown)


samples/word/50-document/insert-external-document.yaml
```diff
@@ -71,7 +89,12 @@ template:
                 <br>
                 <p>Insert the body text from the selected document.</p>
                 <button id="insert-document" class="ms-Button">
-            <span class="ms-Button-label">Insert document</span>
+          <span class="ms-Button-label">Insert document (no settings)</span>
+          </button>
+                <br>
+                <p>Insert the body text with chosen settings applied from the selected document.</p>
```
AlexJerabek, line 95: Since the user isn't "choosing" the settings, I'd omit that word.
```suggestion
                <p>Insert the body text with settings applied from the selected document.</p>
```

samples/word/50-document/insert-external-document.yaml
```diff
@@ -71,7 +89,12 @@ template:
                 <br>
```
AlexJerabek, line 89: With the expanded scope, consider changing the preceding label from
"Select a Word document to copy its body text into the current document."
to
"Select a Word document."

(I don't think you need to spell out why you're selecting a file, since that's mentioned two other places in the sample.

AlexJerabek (approved):
Looks good and works well. Just a couple suggestions.

PR #852, opened 2024-02-09
[Word] (document) Compare documents

ElizabethSamuel-MSFT:
In response to a request via verbatim


samples/word/99-preview-apis/compare-documents.yaml
```diff
@@ -0,0 +1,78 @@
...
+          } catch (error) {
+            // Note: In a production add-in, you'd want to notify the user through your add-in's UI.
+            console.error(error);
+          }
+        }
+    language: typescript
+template:
+    content: |-
+        <section class="ms-font-m">
+            <p>This sample shows how to compare two documents (the current one and a specified external one).</p>
```
AlexJerabek, line 42:
```suggestion
            <p>This sample shows how to compare two documents: the current one and a specified external one.</p>
```

samples/word/99-preview-apis/compare-documents.yaml
```diff
@@ -0,0 +1,78 @@
...
+        </section>
+
+        <section class="samples ms-font-m">
+            <h3>Try it out</h3>
+            <label>File (with absolute path): </label>
```
AlexJerabek, line 47: Could this be a file picker instead?

AlexJerabek (approved):
Approving, with the understanding that there will be another pass to check the comparison logic.

samples/word/99-preview-apis/compare-documents.yaml
ElizabethSamuel-MSFT: @AlexJerabek I'll look into that

ElizabethSamuel-MSFT: Looks like, for security reasons, can't get absolute path from the browser. The filename is real but nothing else - you get C:\fakepath\Document.docx

PR #861, opened 2024-02-28
[Word] (custom XML part) Also map namespace example

ElizabethSamuel-MSFT:


config/build.documentation.ts
```diff
@@ -161,13 +161,13 @@ function getExtractedDataFromSnippet(
...
-            const editedFilename = filename.substr(filename.lastIndexOf('samples')).replace(/\\/g, '/');
-            text = '// Link to full sample: https://raw.githubusercontent.com/OfficeDev/office-js-snippets/prod/' + editedFilename + '\n';
+            const editedFilename = filename.substring(filename.lastIndexOf('samples')).replace(/\\/g, '/');
+            text = '// Link to full sample: https://raw.githubusercontent.com/OfficeDev/office-js-snippets/prod/' + editedFilename + '\n\n';
 
             text += indented
                 .map(line => {
-                    if (line.substr(0, whitespaceCountOnFirstLine).trim().length === 0) {
-                        return line.substr(whitespaceCountOnFirstLine);
+                    if (line.substring(0, whitespaceCountOnFirstLine).trim().length === 0) {
```
AlexJerabek, line 169: `substr` and `substring` have different behaviors. The former treats the second parameter as a length, whereas the latter uses the second parameter as an end index.

ElizabethSamuel-MSFT: It seems like [substr is deprecated](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr). I'll update how we're using substring properly.

ElizabethSamuel-MSFT: Since the swap didn't lead to any spacing changes, I'm thinking to leave this as is.

AlexJerabek: Maybe that implies this line is never being hit? Not important for this PR, but strange that there were no consequences.

AlexJerabek (approved):

config/build.documentation.ts
ElizabethSamuel-MSFT: I made some changes locally and it does get hit. There were no changes because we're returning the same value as the start index in both cases. I could update the if condition to use the index instead of the length but the outcome is the same.

PR #885, opened 2024-05-29
[Excel] (Workbook) Add sample for built-in Excel functions

alison-mk:
Adding samples from [Call built-in Excel worksheet functions](https://learn.microsoft.com/en-us/office/dev/add-ins/excel/excel-add-ins-worksheet-functions?view=excel-js-preview) to Script Lab, based on customer request.


AlexJerabek (approved):
Looks good! Thanks for responding to customer requests and adding this.

PR #921, opened 2024-07-08
[Word] (list) Add manage-style snippet (preview)

ElizabethSamuel-MSFT:


samples/word/99-preview-apis/manage-list-styles.yaml
```diff
@@ -0,0 +1,112 @@
+id: word-lists-manage-styles
+name: Manage list styles
```
AlexJerabek, line 2: "Manage" seems to oversell the API capabilities. Consider "Get" instead.
```suggestion
name: Get list styles
```

samples/word/99-preview-apis/manage-list-styles.yaml
```diff
@@ -0,0 +1,112 @@
...
+    content: |
+        $("#count").on("click", () => tryCatch(getCount));
+        $("#properties").on("click", () => tryCatch(getProperties));
+
+        async function getCount() {
+          // Gets the available list styles stored with the document.
+          await Word.run(async (context) => {
+            const styles: Word.StyleCollection = context.document.getStyles();
+            const count = styles.getCount();
+            styles.load();
```
AlexJerabek, line 18: Here and elsewhere, let's avoid loading more properties than necessary to show best practices.

samples/word/99-preview-apis/manage-list-styles.yaml
```diff
@@ -0,0 +1,112 @@
...
+
+            for (let i = 0; i <= count.value; i++) {
+              if (styles.items[i] && styles.items[i].type == "List") {
+                console.log(`List style name: ${styles.items[i].nameLocal}`, styles.items[i]);
+              }
+            }
+          });
+        }
+
+        async function getProperties() {
```
AlexJerabek, line 29: The only properties shown are the list levels. Can we rename this function (and the buttons and comments) to reflect that?

AlexJerabek (changes requested):
Looks good! Just a couple comments, though if you take them, I'd like to double-check things before merging.

samples/word/99-preview-apis/manage-list-styles.yaml
ElizabethSamuel-MSFT: @AlexJerabek Yeah, I have some further changes planned but I can rename (and update descriptions, etc.) accordingly when that time comes.

samples/word/99-preview-apis/manage-list-styles.yaml
ElizabethSamuel-MSFT: I'm loading everything for now so users can see all the properties in the log. Maybe I'll add a code comment to that effect, which will explain why we're not strictly following best practice.

samples/word/99-preview-apis/manage-list-styles.yaml
ElizabethSamuel-MSFT: Right now, we're logging all the properties of the list style along with the list levels so the name, etc. is appropriate.

AlexJerabek (approved):

PR #957, opened 2025-01-16
[Excel] (Data types) Switch to licensed images

alison-mk:
The new images are from our assets library and licensed for general Microsoft use. 

To test the new images in this PR, you can use this link format in Script Lab: 

https://github.com/OfficeDev/office-js-snippets/blob/alison-mk-data-types-images/.github/images/chili-peppers.jpg?raw=true


davidchesnut (approved):
Looks good, thanks for updating these!

alison-mk:
Note: I checked and this update doesn't require running yarn start.

PR #961, opened 2025-01-31
[Admin] main -> prod

ElizabethSamuel-MSFT:


AlexJerabek (approved):
Looks good to me. Should we automate this publishing too?

Edit: So we are! Nevermind.

PR #968, opened 2025-03-19
[Outlook] (item multi-select) Create code snippet for loadItemByIdAsync method

samantharamon:


samples/outlook/90-other-item-apis/get-loaded-message-properties.yaml
```diff
@@ -0,0 +1,154 @@
...
+
+            run();
+          }
+        });
+
+        function run() {
+          // Clear the list of previously selected messages, if any.
+          clearList(list);
+
+          // Get the subject line and sender's email address of each selected message and log it to a list in the task pane.
```
ElizabethSamuel-MSFT, line 34:
```suggestion
          // Get the subject line and sender's email address of each selected message and log them to a list in the task pane.
```

ElizabethSamuel-MSFT (approved):

PR #971, opened 2025-03-25
[All hosts] Convert jQuery to JavaScript

samantharamon:
Converts jQuery instances that weren't picked up by the bulk conversion tool (see https://github.com/OfficeDev/office-js-snippets/pull/970). Updated samples have been tested.

Custom function samples will be addressed separately.

### DIFF
--- a/samples/excel/99-just-for-fun/gradient.yaml
+++ b/samples/excel/99-just-for-fun/gradient.yaml
@@ -8,8 +8,6 @@ api_set:
     ExcelApi: '1.4'
 script:
     content: |-
-        // initializeColorPickers();
-
         // Set up the click handler.
         document.getElementById("draw-gradient").addEventListener("click", drawGradient);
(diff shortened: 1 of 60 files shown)


samples/excel/99-just-for-fun/gradient.yaml
```diff
@@ -8,26 +8,27 @@ api_set:
     ExcelApi: '1.4'
 script:
     content: |-
-        initializeColorPickers();
+        // initializeColorPickers();
```
AlexJerabek, line 11: Since we removed this function, should we just delete this line of code?

AlexJerabek (approved):

samples/excel/99-just-for-fun/gradient.yaml
samantharamon: Thanks for catching this. I had commented it out during testing and forgot to remove it.

PR #973, opened 2025-04-01
[Excel] (Range) Add cell control and notes snippets

alison-mk:


samples/excel/42-range/range-cell-control.yaml
```diff
@@ -0,0 +1,120 @@
...
+          };
+          await context.sync();
+        });
+      }
+      
+      // Remove checkboxes from the table.
+      async function removeCheckboxes() {
+        await Excel.run(async (context) => {
+          const sheet = context.workbook.worksheets.getActiveWorksheet();
+          let range = sheet.getRange("B2:B9");
```
jakobpn, line 31: Consider a reference to the column rather than the B2:B9 range.

samples/excel/42-range/range-cell-control.yaml
```diff
@@ -0,0 +1,120 @@
...
+    content: |-
+      document.getElementById("setup").addEventListener("click", () => tryCatch(setup));
+      document.getElementById("add-checkboxes").addEventListener("click", () => tryCatch(addCheckboxes));
+      document.getElementById("remove-checkboxes").addEventListener("click", () => tryCatch(removeCheckboxes));
+      
+      // Add checkboxes to the table.
+      async function addCheckboxes() {
+        await Excel.run(async (context) => {
+          const sheet = context.workbook.worksheets.getActiveWorksheet();
+          let range = sheet.getRange("B2:B9");
```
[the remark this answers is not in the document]
alison-mk: Thanks for the review, Jacob! I incorporated your feedback by having the sample replace all boolean values in the table with checkboxes.

samples/excel/42-range/range-cell-control.yaml
```diff
@@ -0,0 +1,120 @@
+order: 23
+id: excel-cell-controls
+name: Checkboxes
+description: This sample shows how to add and remove checkboxes from a table.
```
AlexJerabek, line 4: Is there a way to edit a checkbox?

AlexJerabek (approved):
These look good to me!

samples/excel/42-range/range-cell-control.yaml
```diff
@@ -0,0 +1,120 @@
...
+    content: |-
+        document.getElementById("setup").addEventListener("click", () => tryCatch(setup));
+        document.getElementById("add-checkboxes").addEventListener("click", () => tryCatch(addCheckboxes));
+        document.getElementById("remove-checkboxes").addEventListener("click", () => tryCatch(removeCheckboxes));
+
+        // Add checkboxes to the table.
+        async function addCheckboxes() {
+          await Excel.run(async (context) => {
+            const sheet = context.workbook.worksheets.getActiveWorksheet();
+            let range = sheet.getRange("FruitTable");
```
jakobpn, line 18: The code will add the checkbox control (formatting) to the entire range for the table. I doubt that is what we want to demonstrate?

samples/excel/42-range/range-cell-control.yaml
```diff
@@ -0,0 +1,120 @@
...
+            };
+            await context.sync();
+          });
+        }
+
+        // Remove checkboxes from the table.
+        async function removeCheckboxes() {
+          await Excel.run(async (context) => {
+            const sheet = context.workbook.worksheets.getActiveWorksheet();
+            let range = sheet.getRange("B2:B9");
```
jakobpn, line 31: Would be better to show how to reference the entire column.

samples/excel/42-range/range-cell-control.yaml
AlexJerabek: So why didn't the whole table display checkboxes then?

jakobpn: Because the cells in the table header and first column have string values. If you change that to TRUE or FALSE you will see a checkbox show up.

alison-mk: That is exactly what I'm trying to demonstrate. All boolean values get changed to checkboxes in the selected range.

alison-mk: I updated the snippet to get the Analysis column: 

`const range = sheet.getRange("FruitTable").getColumn(1)`

jakobpn: Excellent! That is a lot better.

samples/excel/42-range/range-cell-control.yaml
```diff
@@ -0,0 +1,163 @@
...
+        document.getElementById("remove-checkboxes").addEventListener("click", () => tryCatch(removeCheckboxes));
+        document.getElementById("remove-content").addEventListener("click", () => tryCatch(removeContent));
+
+        async function addCheckboxes() {
+          // Add checkboxes to the table.
+          await Excel.run(async (context) => {
+            const sheet = context.workbook.worksheets.getActiveWorksheet();
+
+            // Get the second column in the table.
+            const range = sheet.getRange("FruitTable").getColumn(1);
```
AlexJerabek, line 22: Consider not including the header in that column.
```suggestion
            const range = sheet.tables.getItem("FruitTable").columns.getItem("Analysis").getDataBodyRange();
```

samples/excel/42-range/range-cell-control.yaml
```diff
@@ -0,0 +1,163 @@
...
+            fruitTable.rows.add(null /*add at the end*/, [
+              ["Apple", "TRUE"],
+              ["Kale", "FALSE"],
+              ["Orange", "TRUE"],
+              ["Pear", "TRUE"],
+              ["Mango", "TRUE"],
+              ["Dog", "FALSE"],
+              ["Strawberry", "TRUE"],
+              ["Chair", "FALSE"]
+            ]);
```
AlexJerabek, line 93: There was a comment thread between @alison-mk and @jakobpn that seemed to center around should we show the behavior of setting a checkbox on something that's not a boolean. @alison-mk's original code had everything in the table get the checkbox treatment, which showed a range of behavior, but wouldn't be a good pattern for developers to copy. How about, as a compromise, we add a row here.
```suggestion
              ["Chair", "FALSE"],
              ["Tomato", "UNKNOWN"] /* Only boolean values will render as checkboxes. */
            ]);
```

AlexJerabek (approved):
Approving with a couple suggestions.

samples/excel/42-range/range-cell-control.yaml
alison-mk: Great idea, thanks Alex! And lol at the tomato choice.

AlexJerabek (approved):

PR #974, opened 2025-04-04
[Word] Add snippets for APIs re pages, panes, windows

ElizabethSamuel-MSFT:
Will do object mapping later


samples/word/35-ranges/get-pages.yaml
```diff
@@ -0,0 +1,244 @@
+order: 4
+id: word-ranges-get-pages
+name: 'Work with pages, panes, and windows'
+description: 'Shows how to work with pages, panes, and windows APIs.'
```
AlexJerabek, line 4: I think that we're working with APIs is already implied since we're in Script Lab. 
```suggestion
description: 'Shows how to work with pages, panes, and windows.'
```

samples/word/35-ranges/get-pages.yaml
```diff
@@ -0,0 +1,244 @@
...
+              await context.sync();
+
+              console.log("Text of that page in the selection:", range.text);
+            }
+          });
+        }
+
+        async function getPagesOfThirdParagraph() {
+          await Word.run(async (context) => {
+            // Gets the pages that the third paragraph is found on.
```
AlexJerabek, line 43: The dangling "on" seemed off, but I think this is the most concise way to articulate the relationships.
```suggestion
            // Gets the pages that contain the third paragraph.
```

samples/word/35-ranges/get-pages.yaml
```diff
@@ -0,0 +1,244 @@
...
+            paragraphs.load();
+            await context.sync();
+
+            const paraThree = paragraphs.items[2];
+            const rangeOfParagraph = paraThree.getRange();
+            await context.sync();
+
+            const pages: Word.PageCollection = rangeOfParagraph.pages;
+            pages.load();
+            await context.sync();
```
AlexJerabek, line 54: Do we really need three roundtrips to the document before even getting to the API in question? If so, that seems like a product flaw.

samples/word/35-ranges/get-pages.yaml
```diff
@@ -0,0 +1,244 @@
...
+            }
+          });
+        }
+
+        async function getPagesEnclosingViewport() {
+          await Word.run(async (context) => {
+            // Gets the pages enclosing the viewport.
+
+            // Get the active window.
+            let activeWindow = context.document.activeWindow;
```
AlexJerabek, line 78: We switched from let to const partway through the snippet. Let's use const here and elsewhere.
```suggestion
            const activeWindow = context.document.activeWindow;
```

samples/word/35-ranges/get-pages.yaml
```diff
@@ -0,0 +1,244 @@
...
+            activeWindow.load();
+            await context.sync();
+
+            // Get the active pane.
+            let activePane = activeWindow.activePane;
+            activePane.load();
+            await context.sync();
+            // Get pages enclosing the viewport.
```
AlexJerabek, line 86: I find adding a blank line after syncs helps with readability.
```suggestion
            await context.sync();
            
            // Get pages enclosing the viewport.
```

samples/word/35-ranges/get-pages.yaml
```diff
@@ -0,0 +1,244 @@
...
+            const pages: Word.PageCollection = context.document.getSelection().pages;
+            pages.load();
+            await context.sync();
+
+            console.log(pages);
+            // Log info for pages included in selection.
+            for (let i = 0; i < pages.items.length; i++) {
+              let page = pages.items[i];
+              page.load();
+              await context.sync();
```
AlexJerabek, line 29: We should avoid doing load/sync in a loop. It's a bad pattern. This article has some tips to avoid that: https://learn.microsoft.com/en-us/office/dev/add-ins/concepts/correlated-objects-pattern

samples/word/35-ranges/get-pages.yaml
```diff
@@ -0,0 +1,244 @@
...
+          } catch (error) {
+            // Note: In a production add-in, you'd want to notify the user through your add-in's UI.
+            console.error(error);
+          }
+        }
+    language: typescript
+template:
+    content: |-
+        <section class="ms-Fabric ms-font-m">
+            This sample demonstrates how to work with the pages, panes, and windows APIs.
```
AlexJerabek, line 196:
```suggestion
            This sample demonstrates how to work with pages, panes, and windows.
```

AlexJerabek (changes requested):

AlexJerabek (approved):
Looks good!

AlexJerabek (approved):

PR #975, opened 2025-04-23
Remapping cell control from a class to a type

AlexJerabek:
I think some recent changes to the doc generation may have unlinked this type alias snippet. Here's the mapping fix.


alison-mk (approved):
Thanks Alex!

PR #986, opened 2025-06-11
[PowerPoint] (tables) Work with text runs

ElizabethSamuel-MSFT:


samples/powerpoint/shapes/add-modify-tables.yaml
```diff
@@ -201,6 +202,23 @@ script:
...
 
+        async function tableTextRuns() {
+          // Specifies the text runs of the cells in a table.
+          await PowerPoint.run(async (context) => {
+            const shapes = context.presentation.getSelectedSlides().getItemAt(0).shapes;
+
+            // Add a table, specifying text runs.
+            shapes.addTable(3, 4, {
+              specificCellProperties: [
+                [{ textRuns: [{ text: "Bold text", font: { bold: true } }] }, { textRuns: [{ text: "Bold text", font: { bold: true } }] }, { textRuns: [{ text: "Bold text", font: { bold: true } }] }, { textRuns: [{ text: "Bold text", font: { bold: true } }] }],
```
akrantz: You might consider different choices about the formatted text for each cell. Also, it doesn't make sense to use a single text run for formatted text. The intent is to have multiple different formatting within the cell text. Otherwise, you can just set the cell formatting directly and use the text property to set the text (whose formatting defaults to the cell formatting). 

You could even show setting for example a cell to have bold text by default, and then specifying italic for some range of text within the cell. That would show that the bold is inherited from the cell formatting and then the run with italic is set. That would look like for a single specificCellProperties:

```
{ bold: true, textRuns: [{ text: "Here is " }, {text: "italic text", italic: true}, {text: "while the other text is not"}]}
```

akrantz: For a cell which doesn't have a mixture of formatting for the text, you can show this as:

```
{ italic: true, text: "this text is underlined" }
```

akrantz (changes requested):
See comments.

akrantz (approved):

PR #993, opened 2025-07-18
[Excel] (Shape) Add get active shape image snippet

alison-mk:
This snippet shows how to get an image of the active shape and return it to the task pane.

This PR is ready to merge.


samples/excel/44-shape/shape-get-active.yaml
```diff
@@ -0,0 +1,175 @@
...
+            shape.textFrame.textRange.font.size = 13;
+
+            // Get chart positioning information to place the shape correctly.
+            const chart = sheet.charts.getItemAt(0);
+            chart.load("left, top");
+            chart.plotArea.load("width");
+            await context.sync();
+
+            // Position the shape above the chart.
+            shape.left = chart.left + chart.plotArea.width / 2 - 30;
```
AlexJerabek: Just checking that the order of operations in this math reflects what you want.

samples/excel/44-shape/shape-get-active.yaml
```diff
@@ -0,0 +1,175 @@
...
+          });
+        }
+
+        async function getActiveShape() {
+          // This method gets the active shape and displays it as an image in the task pane.
+          await Excel.run(async (context) => {
+            // Get the currently active shape, if any.
+            const activeShape = context.workbook.getActiveShapeOrNullObject();
+            
+            if (activeShape == null) {
```
AlexJerabek: Consider rewriting this to be `if (activeShape)` and leading with the positive case.

samples/excel/44-shape/shape-get-active.yaml
alison-mk: Good catch. Adjusted.

AlexJerabek (approved):

AlexJerabek (approved):

PR #1000, opened 2025-09-27
[Excel] Map existing snippets to ref docs

alison-mk:


snippet-extractor-output/snippets.yaml
```diff
@@ -445,6 +445,246 @@ Excel.CalculationType:enum:
       context.application.calculate(Excel.CalculationType.recalculate);
       await context.sync();
     });
+Excel.CardLayoutSection:type:
```
AlexJerabek, line 448: I'm worried this snippet is so large, it'll be hard to find the relevant section. Is all this context necessary, or it is an artifact of the way the snippet mapping works?

alison-mk: Good point. I updated the PR and mapped CardLayoutSection to a different, shorter snippet. I also mapped EntityPropertyType and EntityViewLayouts (others in this PR) to the shorter snippet.

snippet-extractor-output/snippets.yaml
```diff
@@ -445,6 +445,141 @@ Excel.CalculationType:enum:
...
+  - >-
+    // Link to full sample:
+    https://raw.githubusercontent.com/OfficeDev/office-js-snippets/prod/samples/excel/20-data-types/data-types-entity-attribution.yaml
+
+
+    function makeProductEntity(productID: number, productName: string, product?:
```
AlexJerabek, line 454: Where is the actual CardLayoutSection in this sample? It's not obvious (to me) from skimming the code.

AlexJerabek (approved):
Thanks @alison-mk! I think these are good snippets to use. My one thought is that it's not clear where the interface is used in some of these samples.  We often make the type explicit (e.g. ` const entity: Excel.EntityCellValue = ...`) so that people understand how the code sample links to the reference page, without needing to understand the object model (e.g. the DV sample requires you know `rankingRange.dataValidation.errorAlert` is of type `DataValidationErrorAlert`). Consider adding explicit typing or comments to clear this up.

snippet-extractor-output/snippets.yaml
alison-mk: Great feedback! I added the explicit types as inline comments. Please take a look and let me know what you think.

AlexJerabek (approved):
Looks good!

PR #1090, opened 2026-05-07
[excel] (Range) Merge precedent and dependent samples

AlexJerabek:
The three snippets about this topic were very similar. This PR merge them into a single snippet for better discoverability. 

This PR also updates the tsconfig files to avoid some VSCode warnings.


samples/excel/42-range/precedents-and-dependents.yaml
```diff
@@ -0,0 +1,309 @@
...
+            console.error(error);
+          }
+        }
+    language: typescript
+template:
+    content: |-
+        <section>
+          <p>This sample shows how to find and highlight <b>precedents</b> and <b>dependents</b> of a selected cell.</p>
+          <ul>
+            <li><b>Precedents</b> are cells that a formula references (what the formula <i>depends on</i>).</li>
```
davidchesnut, line 233: It would be helpful to add the color coding here (yellow and orange) to make it a bit more obvious when clicking in the spreadsheet as to what is shown.

davidchesnut (approved):
One suggestion added. Also wondering if it should show an error when you select a cell without dependents or precedents? Or should it show something like "none" in the console.

Otherwise I like the combining!

AlexJerabek:
> One suggestion added. Also wondering if it should show an error when you select a cell without dependents or precedents? Or should it show something like "none" in the console.
> 
> Otherwise I like the combining!

Thanks @davidchesnut! That's a good idea - I'll add it.

AlexJerabek:
@davidchesnut, so it turns out the precedent/dependent APIs throw errors by design when they have no precedents or dependents. There's not really an empty check that wouldn't overcomplicate the sample.

davidchesnut (approved):

PR #1092, opened 2026-05-15
[word] (Coauthoring) Add coauthoring samples for Word

AlexJerabek:
This PR adds a few samples for the desktop-only Word Coauthoring APIs. Note that there appears to be a product bug in the Lock sample, so none of those APIs are mapped. When confirmed or mitigated, I'll update the sample.


samples/word/05-coauthoring/manage-coauthoring-locks.yaml
```diff
@@ -0,0 +1,215 @@
...
+            await context.sync();
+
+            if (!coauthoring.canCoauthor) {
+              console.warn("This document doesn't support coauthoring. Save it to OneDrive or SharePoint and try again.");
+              return;
+            }
+
+            const lock: Word.CoauthoringLock = coauthoring.locks.add({
+              type: Word.CoauthoringLockType.reservation
+            });
```
ElizabethSamuel-MSFT: I'm able to get a reservation lock by using the default behavior. It might be a bug that the param values aren't being honored or recognized.

```suggestion
            const lock: Word.CoauthoringLock = coauthoring.locks.add();
```

AlexJerabek: Thanks @ElizabethSamuel-MSFT. I filed an internal bug and rewrote the sample. I'm confirming the fixes work and updating the test suite.

ElizabethSamuel-MSFT (approved):

PR #1099, opened 2026-07-23
[excel] (Data Types) Update data types sample to show basic functionality more clearly

AlexJerabek:


samples/excel/20-data-types/data-types-basic-types.yaml
```diff
@@ -109,226 +210,72 @@ script:
...
         </section>
         <section>
             <h3>Try it out</h3>
-            <p>Add basic types to the worksheet. Select the icon in each cell to see a data type card with metadata.</p>
-            <button id="set-basic-types-with-metadata">Create basic types with metadata</button>
-        </section>
-        <section>
-            <p>Add a basic type with referenced values. Open the data type card to see other values in the worksheet that this basic type references&mdash;it references cells <strong>A1</strong>, <strong>A2</strong>, <strong>A3</strong>, and <strong>A4</strong>.</p>
-            <button id="set-referenced-values">Create basic type with referenced values</button>
+            <p>Add ticket totals that carry metadata. The cell shows the total. Open the data type card to reveal the purchased items that make up that total.</p>
```
alison-mk: I'm not seeing a data type card icon in Excel on Mac or Excel on the web, which makes the direction to "Open the data type card" confusing.  

Mac: 
Image

Web: 
Image

I can open the data type card if I right-click and then choose Show Data Type Card: 
Image

Maybe the guidance should clarify this step? 

I see that the data type icons being off by default is explained in the code: `For basic types with metadata, unlike entity values, the cell shows the value itself rather than an icon (icons are off by default).` I think we should consider adding mention of this to the HTML content—the sample felt broken to me until I figured out what was going on.

alison-mk: Unclear why my Excel on the web is showing the Euro number sign. Potential bug?

alison-mk (approved):
Thanks for your patience! Approved with one suggestion, and one potential bug.

samples/excel/20-data-types/data-types-basic-types.yaml
AlexJerabek: I'm unable to repro the Euro sign bug, which leads me to believe there's some underlying number format parsing difference between the data type APIs and the range APIs. Copilot's analysis is:

> That distinction matches the screenshot:
A2:A4 show $ but still use the reviewer’s locale for separators, so they render like $ 74,25.
F1:F3 show €, which means Excel treated those cells as regular currency-formatted cells and substituted the viewer’s local currency symbol.
>
> So the likely behavior is:
rich data-type values preserved the literal dollar marker from the value format,
regular cell formatting localized the currency symbol to the user’s regional settings.
>
> That’s why you get mixed output even though both places started from "$* #,##0.00".

I'll forward this along to the product team, since I'd have to overwork the formatting in the sample to hide this.

**Desktop**
image

**Web**
image

AlexJerabek:
Thanks @alison-mk! I added a line to the HTML explaining how to open the data type card. I think that'll make it clearer what the sample is doing without changing the in-Excel UI.

alison-mk (approved):

